Issue #26, opened 2023-11-27
Support DALL-E 3 text-to-image API


avavilau:

VolhaBazhkova:
DALL-E 3 works fine on staging server, also statistics is gathered in Grafana. Please note, that there is a limit of 6 pic in a minute.

PR #30, opened 2023-12-05
feat: support DALL-E 3 (#26)

vladisavvv:


vladisavvv:
/deploy-review
>Deployment status: success

aidial_adapter_openai/app.py
```diff
@@ -26,6 +31,20 @@
 model_aliases: Dict[str, str] = json.loads(os.getenv("MODEL_ALIASES", "{}"))
 azure_api_version = os.getenv("AZURE_API_VERSION", "2023-03-15-preview")
 
+dial_use_file_storage = os.getenv("DIAL_USE_FILE_STORAGE", False)
```
adubovik: You need to parse the string to a boolean.

Otherwise, if `DIAL_USER_FILE_STORAGE=False`, then the condition `if dial_use_file_storage:` will still be successful.

You could use a helper like this:

```python
def bool_env_var(name: str, default: bool = False) -> bool:
    return (os.getenv(name, str(default)).lower() == 'true')
```

aidial_adapter_openai/images.py
```diff
@@ -0,0 +1,134 @@
...
+        ) as response:
+            status_code = response.status
+
+            data = await response.json()
+
+            if status_code == 200:
+                return data
+            else:
+                return JSONResponse(content=data, status_code=status_code)
+
```
adubovik, line 37: It's would be more straighforward to raise an exception here instead of returning JSONResponse. There is a FastAPI exception handler, which will turn it eventually into a proper JSONResponse.

aidial_adapter_openai/images.py
```diff
@@ -0,0 +1,134 @@
...
+
+def build_custom_content(base64_image: str, revised_prompt: str) -> Any:
+    return {
+        "custom_content": {
+            "attachments": [
+                {"title": "Image", "type": "image/png", "data": base64_image},
```
adubovik: Swap the order of attachments as discussed

aidial_adapter_openai/images.py
```diff
@@ -0,0 +1,134 @@
...
+async def generate_stream(
+    id: str, created: str, custom_content: Any
+) -> AsyncGenerator[Any, Any]:
+    yield chunk_format(
+        build_chunk(id, None, {"role": "assistant"}, created, True)
+    )
+
+    yield chunk_format(build_chunk(id, None, custom_content, created, True))
+
+    yield chunk_format(
```
adubovik, line 59: You could save a few lines of code if you pull `usage` into to a dedicated variable and then use it in `build_chunk`, so that it will be formatted as a single line.

aidial_adapter_openai/images.py
```diff
@@ -0,0 +1,134 @@
...
+                "prompt_tokens": 1,
+                "total_tokens": 1,
+            },
+        )
+    )
+
+    yield END_CHUNK
+
+
+def validate_request_data(data: Any):
```
adubovik: 1. It's more practical to combine validations like this one with actual extraction of the thing whose existence you are trying to guarantee by this validation. Here is it's the content of the last message. So I suggest to rewrite it as `def get_user_prompt(data) -> str`.
2. Additionally check that content isn't `None`, because it's optional.

aidial_adapter_openai/images.py
```diff
@@ -0,0 +1,134 @@
...
+
+    custom_content = build_custom_content(base64_image, revised_prompt)
+
+    if file_storage is not None:
+        file_metadata = await upload_base64_file(
+            file_storage, base64_image, "image/png"
+        )
+        image_url = file_metadata["path"] + "/" + file_metadata["name"]
+
+        del custom_content["custom_content"]["attachments"][0]["data"]
```
adubovik: It depends on how `build_custom_content` is implemented. Only one image is expected and it must be in the first attachment.

I recommend a helper which checks every attachment if it has type "image/*" and a "data" field and save only those to the storage.

aidial_adapter_openai/images.py
```diff
@@ -0,0 +1,134 @@
...
+    if not is_stream:
+        return JSONResponse(
+            content=build_chunk(
+                id,
+                "stop",
+                {**custom_content, "role": "assistant"},
+                created,
+                False,
+                usage={
+                    "completion_tokens": 0,
```
adubovik: Maybe worth to factor it out into a constant `IMAGE_GENERATION_USAGE` since it's used twice already in the module.

aidial_adapter_openai/images.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+async def text_to_image_chat_completion(
+    data: Any,
+    upstream_endpoint: str,
+    api_key: str,
+    is_stream: bool,
+    file_storage: Optional[FileStorage],
+) -> Response:
+    validate_request_data(data)
```
adubovik: What's worth to validate in the request is that `n=1`.

Alternatively, support `n>1`, which is especially useful for image generation. Typically a user wants a range of images to chose the best from.

aidial_adapter_openai/app.py
```diff
@@ -26,6 +31,20 @@
...
+
+file_storage = None
+if dial_use_file_storage:
+    dial_url = os.getenv("DIAL_URL")
```
adubovik, line 40: Document the new env vars in `README.md`

aidial_adapter_openai/images.py
vladisavvv: do you mean `from fastapi import HTTPException`? Unfortunately it wraps the content in a "detail" block and I don't know how to get around it

aidial_adapter_openai/images.py
vladisavvv: agree, swapped

aidial_adapter_openai/images.py
vladisavvv: agree, created `IMG_USAGE` constant

aidial_adapter_openai/images.py
vladisavvv: agree, rewrote this

aidial_adapter_openai/images.py
vladisavvv: I don't really like these lines either, but we don't plan to put more than one picture in one answer. Even after `n` starts to be supported by `azure` it will be a separate choice in the response.

aidial_adapter_openai/images.py
vladisavvv: yes, sure, created IMG_USAGE

aidial_adapter_openai/app.py
vladisavvv: fixed, thank you

aidial_adapter_openai/app.py
vladisavvv: added

aidial_adapter_openai/images.py
vladisavvv: added the validation. Azure currently support only n=1 in the request :( So, our users will have almost no benefits from a batch request to us and can simulate a batch request themselves

aidial_adapter_openai/images.py
adubovik: Still it's fragile - change in one place will require a change in another.

`build_custom_content` is leaky abstraction.
If we could easily seal the leak, we should do it.

aidial_adapter_openai/images.py
adubovik: You could raise `aidial_adapter_openai.utils.exceptions.HTTPException`.

Alternatively, did you try to use OpenAI lib to do the requests?

```python
response = openai.Image.create(
    prompt='USER_PROMPT_GOES_HERE',
    size='1024x1024',
    n=1
)

image_url = response["data"][0]["url"]
```

This is an example from Azure itself.

I suppose `openai.Image.create` throws `OpenAIException` which you could handle via `handle_exceptions` method (the same way as it done for ChatCompletion calls).

vladisavvv: > aidial_adapter_openai.utils.exceptions.HTTPException 

I don't think this is a good idea due to the fact that we don't really know what the json model will be and how the Azure will decide to change it

vladisavvv: Yes, I tried to use the library from the very beginning, but the problem is that we did not have time to migrate to the 1.X openai python client. In our current version, dall-e 2 api is only supported

aidial_adapter_openai/images.py
vladisavvv: rewrote this part, could you please check?

adubovik (approved):

PR #32, opened 2023-12-07
feat: support tools and tool_choice params (#31)

vladisavvv:


aidial_adapter_openai/utils/request_classifier.py
```diff
@@ -0,0 +1,26 @@
+from typing import Any
+
+
+def is_request_used_functions_or_tools(data: Any):
```
adubovik: 1/ Ungrammatical. Maybe `does_request_use_functions_or_tools`?
2/ I suggest to rename `data` argument to `request`. Data is too generic.

aidial_adapter_openai/utils/request_classifier.py
```diff
@@ -0,0 +1,26 @@
...
+        or "tools" in data
+        or "tool_choice" in data
+    ):
+        return True
+
+    if "messages" not in data:
+        return False
+
+    for message in data["messages"]:
+        role = message.get("role", None)
```
adubovik: Simply `message.get("role")`. It returns None if the key is missing.

vladisavvv: fixed

aidial_adapter_openai/utils/request_classifier.py
vladisavvv: fixed two points

adubovik (approved):

Issue #74, opened 2024-03-04
Support available credentials providers for Azure OpenAI


astsiapanay:
AKS - https://azure.github.io/azure-workload-identity/docs/

Microsoft Entra Workload ID with Azure Kubernetes Service

PR #75, opened 2024-03-04
feat: Support available credentials providers for Azure OpenAI (#74)

astsiapanay:


aidial_adapter_openai/utils/auth.py
```diff
@@ -1,7 +1,41 @@
...
+    if access_token is None or now > access_token.expires_on:
+        access_token = default_credential.get_token(
+            "https://cognitiveservices.azure.com/.default"
+        )
+    return access_token.token
+
+
+async def get_credentials(request: Request) -> tuple[str, str]:
+    api_key = request.headers.get("X-UPSTREAM-KEY")
+    api_type = "azure"
```
vladisavvv: you can move these lines to the else statement to avoid a double-set operation. Or just change this to the next code:
```python
    if api_key is None:
        return "azure_ad", await get_api_key()
    else:
        return "azure", request.headers.get("X-UPSTREAM-KEY")
```

aidial_adapter_openai/utils/auth.py
```diff
@@ -1,7 +1,41 @@
...
+
+EXPIRATION_WINDOW_IN_SEC = 10
+
+
+async def get_api_key() -> str:
+    now = int(time.time()) - EXPIRATION_WINDOW_IN_SEC
+    global access_token
+    if access_token is None or now > access_token.expires_on:
+        access_token = default_credential.get_token(
+            "https://cognitiveservices.azure.com/.default"
```
vladisavvv: let's support env variable for the url

aidial_adapter_openai/utils/auth.py
```diff
@@ -1,7 +1,41 @@
...
+from azure.identity import DefaultAzureCredential
+from fastapi import Request
+from openai import util
+from openai.util import ApiType
 from pydantic import BaseModel
 
+default_credential = DefaultAzureCredential()
+access_token: AccessToken | None = None
+
+EXPIRATION_WINDOW_IN_SEC = 10
```
vladisavvv: let's support env variable for the expiration window

aidial_adapter_openai/utils/auth.py
```diff
@@ -1,7 +1,41 @@
...
+
+
+async def get_api_key() -> str:
+    now = int(time.time()) - EXPIRATION_WINDOW_IN_SEC
+    global access_token
+    if access_token is None or now > access_token.expires_on:
+        access_token = default_credential.get_token(
```
vladisavvv: need to handle the case when we don't have azure auth variables

aidial_adapter_openai/utils/auth.py
astsiapanay: fixed

aidial_adapter_openai/utils/auth.py
astsiapanay: fixed

aidial_adapter_openai/utils/auth.py
astsiapanay: fixed

aidial_adapter_openai/utils/auth.py
astsiapanay: fixed

vladisavvv (approved):

Issue #67, opened 2024-02-19
Migrate to the new default azure api version


vladisavvv:
The current version 2023-03-15-preview will be deprecated. 

The official announcement: https://learn.microsoft.com/en-us/azure/ai-services/openai/api-version-deprecation

<img width="796" alt="image" src="https://github.com/epam/ai-dial-adapter-openai/assets/24621956/dc22a8fe-658a-45da-adb6-d4c11966fe6b">

PR #79, opened 2024-03-11
feat!: reworking api versions logic (#67)

vladisavvv:


aidial_adapter_openai/app.py
```diff
@@ -58,6 +56,16 @@ async def handle_exceptions(call):
         )
 
 
+def get_api_version(request: Request):
+    api_version = request.query_params.get("api-version", "")
+
+    return (
```
Allob: You could just use `get(key[, default])` with default value like this:
```
return api_versions_mapping.get(api_version, api_version)
```

aidial_adapter_openai/app.py
```diff
@@ -172,14 +184,25 @@ async def embedding(deployment_id: str, request: Request):
...
+    api_version = get_api_version(request)
+
+    if api_version == "":
+        raise HTTPException(
+            "Api version is a required query parameter",
+            400,
+            "invalid_request_error",
+        )
+
+    if api_version in api_versions_mapping:
```
Allob: Why do we need this?
`get_api_version` should already do the `api_versions_mapping` transformation.

aidial_adapter_openai/dalle3.py
```diff
@@ -126,7 +127,7 @@ async def chat_completion(
             type="invalid_request_error",
         )
 
-    api_url = upstream_endpoint + "?api-version=2023-12-01-preview"
+    api_url = upstream_endpoint + f"?api-version={api_version}"
```
Allob: This would generally work, but the better way to add a query parameter would be by using functions from urllib.parse:
Something like this:
```
api_url = urlparse(upstream_endpoint)._replace(query=urlencode({"api-version" : api_version})).geturl()
```
This would handle the cases like upstream_endpoint already ending with "?" character.

tests/test_versions.py
```diff
@@ -1,20 +0,0 @@
-from typing import Optional
-
-import pytest
-
-from aidial_adapter_openai.utils.versions import compare_versions
-
-compare_versions_dataset = [
```
Allob, line 7: Do we have an option to have some tests for mapping? For example, by mocking `ChatCompletion().acreate()`?

Allob (dismissed):

tests/test_versions.py
vladisavvv: Unfortunately, we haven't found a nice way to cover system variables with tests. As a temporary solution, I did manual testing.

aidial_adapter_openai/dalle3.py
vladisavvv: Since upstream_endpoint comes from the dial core configuration, I think we can expect that this url has been validated and does not contain extra characters.

aidial_adapter_openai/app.py
vladisavvv: removed, thank you

aidial_adapter_openai/app.py
vladisavvv: fixed, thank you

aidial_adapter_openai/app.py
```diff
@@ -67,11 +71,25 @@ async def chat_completion(deployment_id: str, request: Request):
     api_type, api_key = await get_credentials(request)
 
     upstream_endpoint = request.headers["X-UPSTREAM-ENDPOINT"]
+    api_version = get_api_version(request)
+
+    if api_version == "":
```
Allob: Btw, this check could be moved into get_api_version, since it always needed after get_api_version call.

Allob (dismissed):

aidial_adapter_openai/app.py
vladisavvv: fixed

Allob (approved):

PR #95, opened 2024-04-11
feat: supported Databricks chat completions and embeddings models (#96)

adubovik:
Resolves https://github.com/epam/ai-dial-adapter-openai/issues/96

PR #98, opened 2024-04-15
fix: fixed bug with conflicting "model" fields in the request

adubovik:
When the incoming request has `model` field, it must be overwritten with the `model` field, which we derive from the deployment id.

Fixes #95


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-98.nightly-test.deltixhub.io
>E2E tests status: success

vladisavvv (approved):

PR #113, opened 2024-06-05
feat: add legacy completions API support

### LINKED ISSUE #112: Support legacy completion models
Support ability to connect endpoints which implement legacy `/completions` [API](https://platform.openai.com/docs/api-reference/completions/create).

Make sure 
1. both AzureOpenAI and vanilla OpenAI authentication are supported,
2. streaming and non-streaming modes work

Success criteria: implement NF application via the adapter.

roman-romanov-o:
Closes https://github.com/epam/ai-dial-adapter-openai/issues/112


ai-dial-actions:
Hey there and thank you for opening this pull request! 👋🏼

We require pull request titles to follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/) and it looks like your proposed title needs to be adjusted.

Details:

```
Unknown release type "Feature" found in pull request title "Feature: add legacy completions API support". 

Available types:
 - fix
 - hotfix
 - feat
 - feature
 - chore
```

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-113.nightly-test.deltixhub.io
>E2E tests status: failed

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,102 @@
...
+                },
+                "logprobs": choice["logprobs"],
+            }
+            for choice in item_dict["choices"]
+        ],
+    }
+
+
+async def legacy_completions(
+    data,
```
adubovik: Let's add type hints to the arguments

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,102 @@
...
+        "object": "chat.completion",
+        "id": item_dict["id"],
+        "model": item_dict["model"],
+        "created": item_dict["created"],
+        "usage": item_dict["usage"],
+        "choices": [
+            {
+                "index": choice["index"],
+                "finish_reason": choice["finish_reason"],
+                "message": {
```
adubovik: In the streaming mode the `message` field is called `delta`.

aidial_adapter_openai/app.py
```diff
@@ -43,50 +51,10 @@
 init_telemetry(app, TelemetryConfig())
 configure_loggers()
 
-model_aliases: Dict[str, str] = json.loads(os.getenv("MODEL_ALIASES", "{}"))
```
adubovik, line 46: Let's make the refactoring in a dedicated MR.

aidial_adapter_openai/app.py
```diff
@@ -100,30 +68,40 @@ def get_api_version(request: Request):
 
 @app.post("/openai/deployments/{deployment_id}/chat/completions")
 async def chat_completion(deployment_id: str, request: Request):
+
     data = await parse_body(request)
 
     is_stream = data.get("stream", False)
 
-    api_type, api_key = await get_credentials(request, chat_completions_parser)
+    if deployment_id in LEGACY_COMPLETIONS_DEPLOYMENTS:
```
adubovik: You really don't need to introduce `LEGACY_COMPLETIONS_DEPLOYMENTS` env var, because the format of the upstream endpoint already tells you if this is a chat completions model of a legacy completions model. The parsers do the work of distinguishing one from another.

aidial_adapter_openai/openai_override.py
```diff
@@ -31,11 +35,90 @@ def handle_error_response_wrapper(wrapped, self, args, kwargs):
     raise OpenAIException(*args)
 
 
+# Overriden to handle cases, when server returns valid stream, but wihtout "text/event-stream" header
```
adubovik: wihtout => without

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-113.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/legacy_completions.py
roman-romanov-o: fixed, but adding type hints to data cause adding typing ignore somewhere, because linter thinks that

`data.get('n', 1)` still can be None

aidial_adapter_openai/legacy_completions.py
roman-romanov-o: fixed

aidial_adapter_openai/app.py
roman-romanov-o: fixed

aidial_adapter_openai/openai_override.py
roman-romanov-o: fixed

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,117 @@
...
+from aidial_adapter_openai.types import ChatCompletionRequestData
+from aidial_adapter_openai.utils.auth import get_auth_header
+from aidial_adapter_openai.utils.exceptions import handle_exceptions
+from aidial_adapter_openai.utils.remove_none import remove_none
+from aidial_adapter_openai.utils.sse_stream import to_openai_sse_stream
+from aidial_adapter_openai.utils.streaming import map_stream
+
+
+def convert_completion_to_chat_completion(
+    item: OpenAIObject, role="assistant", is_stream: bool = False
```
adubovik: The defaults here do not make much sense. Why non-streaming is more preferable default than streaming? They are both on the equal footing to me.
Same with the role parameter.

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,117 @@
...
+        "object": "chat.completion",
+        "id": item_dict["id"],
+        "model": item_dict["model"],
+        "created": item_dict["created"],
+        "usage": item_dict.get("usage") or {},
+        "choices": [
+            {
+                "index": choice["index"],
+                "finish_reason": choice["finish_reason"],
+                ("message" if is_stream else "delta"): {
```
adubovik, line 36: It's other way around. Consider reusing the helper:

https://github.com/epam/ai-dial-adapter-openai/blob/ce4a06b308123fa5d3d07ccbc87be4ee252d349d/aidial_adapter_openai/utils/streaming.py#L23-L45

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,117 @@
...
+    return {
+        "object": "chat.completion",
+        "id": item_dict["id"],
+        "model": item_dict["model"],
+        "created": item_dict["created"],
+        "usage": item_dict.get("usage") or {},
+        "choices": [
+            {
+                "index": choice["index"],
+                "finish_reason": choice["finish_reason"],
```
adubovik, line 35: Are you sure "finish_reason" will be present in every single chunk of response? Why not `choice.get("finish_reason")`?

aidial_adapter_openai/utils/remove_none.py
```diff
@@ -0,0 +1,11 @@
+from typing import Any, Dict, Optional
+
+
+def remove_none(d: Dict[str, Optional[Any]]) -> Dict[str, Any]:
+    """
```
adubovik, line 5: IMO, 6 lines of comments for 1 line of actual code is an overkill.

Noone ever is going to read this comment from start to finish. 
One would jump straight away to the line of code.

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,117 @@
...
+
+async def legacy_completions(
+    data: ChatCompletionRequestData,
+    deployment_id: str,
+    is_stream: bool,
+    api_key: str,
+    api_type: str,
+    api_version: str,
+    api_base: str,
+    model=None,
```
adubovik:
```
model: str | None = None,
engine: str | None = None
```

aidial_adapter_openai/utils/parsers.py
```diff
@@ -53,22 +58,32 @@ def parse(self, endpoint: str) -> AzureOpenAIEndpoint | OpenAIEndpoint:
             )
 
         match = re.search(f"(.+?)/{self.name}", endpoint)
-
         if match:
             return OpenAIEndpoint(api_base=match[1])
+        return match
+
+    def is_valid(self, endpoint: str) -> bool:
+        if self._find_match(endpoint):
```
adubovik, line 66: `return self._find_match(endpoint)`

aidial_adapter_openai/utils/parsers.py
```diff
@@ -53,22 +58,32 @@ def parse(self, endpoint: str) -> AzureOpenAIEndpoint | OpenAIEndpoint:
...
+            return True
+        return False
+
+    def parse(self, endpoint: str) -> AzureOpenAIEndpoint | OpenAIEndpoint:
+        if match := self._find_match(endpoint):
```
adubovik, line 69: `return self._find_match(endpoint)`

aidial_adapter_openai/utils/parsers.py
adubovik: `return self._find_match(endpoint) is not None`

aidial_adapter_openai/app.py
```diff
@@ -100,15 +67,37 @@ def get_api_version(request: Request):
...
     data = await parse_body(request)
+    is_stream = bool(data.get("stream", False))
+    api_version = get_api_version(request)
 
-    is_stream = data.get("stream", False)
+    upstream_endpoint = request.headers["X-UPSTREAM-ENDPOINT"]
 
-    api_type, api_key = await get_credentials(request, chat_completions_parser)
+    # If it is not chat completions, but legacy completions
+    if not chat_completions_parser.is_valid(
```
adubovik, line 78: It's better to parse and validate at the same time, so that you don't need to keep in mind that every `parse` should be only called under the branch where `is_valid` is True:

```python
if (endpoint := completions_parser.parse(upstream_endpoint)) is not None:
    ...
```

aidial_adapter_openai/legacy_completions.py
adubovik: `data: dict` is enough :)

I want all of the parameters to be typed to a reasonable extent, which is the case for the rest of the project.

roman-romanov-o: I think, that such `data: dict` type hints are mostly useless, other than preventing some really basic errors.
It's the same as adding `data: object` or `data: Any`.

Even when I know that data is `dict/list/object`, I have no idea at all, how to actually work with this parameter inside of code.

roman-romanov-o: I understand that it's hard to add specific types to all nested objects, but at least we can do it in some places, to make life of developer easier, and prevent him from making `print(data)` to actually understand, what's hidden inside of this dictionary

aidial_adapter_openai/utils/remove_none.py
roman-romanov-o: Currently it doesn't make sense, you're right

But it could make sense to add such `reStructuredText` (or maybe google/numpy type of comments) to each function, so at some point we could generate automatic documentation using some tool like `Sphinx`

aidial_adapter_openai/utils/parsers.py
roman-romanov-o: `self._find_match(endpoint)`could be None, need to check it before returning

aidial_adapter_openai/utils/parsers.py
roman-romanov-o: Done

aidial_adapter_openai/app.py
roman-romanov-o: This would not work with current parsers, since they are raising `DialException` if parsing has not succeeded
I don't really want to change behaviour of this class 

---

Also, another problem, that raises when we send both `/chat/completions` and `/completions` to parser, is that with current `EndpointParser`,

`EndpointParser(name="completions").parse("/some_deployment/chat/completions") will parse successfully, with
`deployment_id = some_deployment/chat`

roman-romanov-o: The approach here, other than I use right now (check that chat completions is not valid in same statement, to filter chat completions from completions) is to create subclass

```python
class CompletionsEndpointParser(EndpointParser):
    ...
```
to exclude `chat/completions` case

aidial_adapter_openai/legacy_completions.py
roman-romanov-o: fixed, removed defaults

aidial_adapter_openai/utils/remove_none.py
adubovik: I agree, but we don't have such plans yet. 

Also, automatic documentation is a good idea for libraries, but this adapter isn't a library but an application on its own.

aidial_adapter_openai/app.py
adubovik: > This would not work with current parsers

The parsers aren't set in stone.

This condition looks so unnatural, because we didn't bother to fix the parsers when `/completions` case was introduced. Conceptually parsers should be disjoint.

PR #105, opened 2024-05-13
feat: migrated to openai==1.33.0

adubovik:
Resolves #71

PR #121, opened 2024-06-18
fix: fixing performance regression

adubovik:
Fixes performance regression caused by https://github.com/epam/ai-dial-adapter-openai/pull/105


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status:

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: success

irinakartun:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: success

irinakartun:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status:

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: failed

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: failed

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: failed

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: success

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/utils/globals.py
```diff
@@ -0,0 +1,17 @@
...
+
+import httpx
+
+from aidial_adapter_openai.constant import (
+    DEFAULT_CONNECTION_LIMITS,
+    DEFAULT_TIMEOUT,
+)
+
+
+@functools.lru_cache(maxsize=1)
```
roman-romanov-o: What do you think, if we make it this way?

```python
class GlobalAsyncClient:
    _client: httpx.AsyncClient | None

    def __init__(self):
        self._client = None

    def set_up_client(self):
        self._client = httpx.AsyncClient(
            timeout=DEFAULT_TIMEOUT,
            limits=DEFAULT_CONNECTION_LIMITS,
            follow_redirects=True,
        )

    async def tear_down_client(self):
        if self._client is None:
            raise Exception("Trying to tear down client before set up!")
        await self._client.aclose()

    @property
    def client(self) -> httpx.AsyncClient:
        if self._client is None:
            raise Exception("Trying to get no setted client!")
        return self._client
```

roman-romanov-o: So then we could make explicit set up and tear down in lifespan:

```python
global_async_client = GlobalAsyncClient()

@asynccontextmanager
async def lifespan(app: FastAPI):
    global_async_client.set_up_client()
    yield
    logger.info("Application shutdown")
    await global_async_client.tear_down_client()
```

aidial_adapter_openai/gpt.py
```diff
@@ -49,27 +50,30 @@ async def gpt_chat_completion(
...
+        {**creds, "api_version": api_version}
     )
 
     response: AsyncStream[ChatCompletionChunk] | ChatCompletion = (
         await call_with_extra_body(client.chat.completions.create, data)
     )
 
     if isinstance(response, AsyncStream):
+
+        def to_dict(chunk: ChatCompletionChunk) -> dict:
```
roman-romanov-o: I you want to use this `to_dict` function to save space, you should consider to move it into `utils`, and re-use it in `databricks` and `mistral` modules, they both have this part


```python
                map_stream(
                    debug_print,
                    map_stream(lambda chunk: chunk.to_dict(), response),
                )
```

It makes sense to do this also perfomance-wise:

Each time we call this `map_stream` twice, we traverse all elements just to make this check for `debug_print`

aidial_adapter_openai/constant.py
```diff
@@ -1,4 +1,9 @@
-from openai import Timeout
+import httpx
 
 # connect timeout and total timeout
-DEFAULT_TIMEOUT = Timeout(600, connect=10)
+DEFAULT_TIMEOUT = httpx.Timeout(600, connect=10)
+
+# Borrowed from openai._constants.DEFAULT_CONNECTION_LIMITS
+DEFAULT_CONNECTION_LIMITS = httpx.Limits(
+    max_connections=1000, max_keepalive_connections=100
```
roman-romanov-o: Minor:

Theoretically we could calculate `max_keepalive_connections` during startup of application, since http polling works that way:

```
When you make several requests to the same host, the Client will reuse the underlying TCP connection, instead of recreating one for every single request.
```

Theoretically, we know exact number of hosts that we we will use, from deployments

adubovik: That's a great idea. Let's leave it for another MR.

aidial_adapter_openai/gpt.py
adubovik: Fixed

aidial_adapter_openai/utils/globals.py
adubovik: And that's not everything - I need to implement a singleton class, because I need to use the instance in other modules too: `.mistral` and `.utils.parsers`.

Too much code for little value. The only thing we gain here is that `lifespan` is more readable and maintainable.
But we loose clarity and maintainability in all other places.

Let's keep it simple.

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-121.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/utils/globals.py
roman-romanov-o: Wait, but if you just make this in `globals.py`

```
global_async_client = GlobalAsyncClient()
```
how is it messing with `mistral` and `utils`?

You're just substituting `get_http_client()` with `global_async_client`

And this `set_up`/`tear_down` is needed just to set up connection after monkey-patch from `opentelemetry` proceeded

---

The main concern I have, is that when instead of object we use this `@functools.cache`:

1. It's really misleading, what we're trying to achieve, because we don't want to cache something, we want one singleton client for all requests
2. I haven't used `functools.cache` for this singleton approach not once in my life. I can't guarantee it don't have some side effects

roman-romanov-o (approved):
I still have this [concerns](https://github.com/epam/ai-dial-adapter-openai/pull/121#discussion_r1647870645) but I guess we can change that easily, so it's not that critical

PR #124, opened 2024-06-25
feat: add legacy completions API support

### LINKED ISSUE #112: Support legacy completion models

roman-romanov-o:
Closes https://github.com/epam/ai-dial-adapter-openai/issues/112


aidial_adapter_openai/app.py
```diff
@@ -125,38 +117,53 @@ async def chat_completion(deployment_id: str, request: Request):
     # Azure and non-Azure deployments.
     # Therefore, we provide the "model" field for all deployments here.
     # The same goes for /embeddings endpoint.
-    data["model"] = deployment_id
+    data["model"] = (
```
roman-romanov-o: Need to add possibility to pass model manually, since we have model names like this: `Test/Test-Model-QA`

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-124.nightly-test.deltixhub.io
>E2E tests status: failed

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-124.nightly-test.deltixhub.io
>E2E tests status: failed

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-124.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/app.py
adubovik: `data.get("model") or deployment_id`

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,87 @@
...
+    messages = data.get("messages", [])
+
+    prompt = messages[-1].get("content") or ""
+
+    if (
+        template := COMPLETION_DEPLOYMENTS_PROMPT_TEMPLATES.get(deployment_id)
+    ) is not None:
+        prompt = template.format(prompt=prompt)
+
+    if len(messages) == 0:
```
adubovik: Move this check before `messages[-1]` is accessed.

aidial_adapter_openai/legacy_completions.py
```diff
@@ -0,0 +1,87 @@
...
+        raise DialException(
+            status_code=422,
+            message="The request doesn't contain any messages",
+            type="invalid_request_error",
+        )
+
+    del data["messages"]
+    del data["stream"]
```
adubovik: No need to remove `stream` field and then add it again.

So you don't need the `is_stream` argument in this method altogether.

aidial_adapter_openai/legacy_completions.py
roman-romanov-o: Fixed

aidial_adapter_openai/legacy_completions.py
roman-romanov-o: Fixed

aidial_adapter_openai/utils/parsers.py
```diff
@@ -52,32 +52,45 @@ def get_client(self, params: OpenAIParams) -> AsyncOpenAI:
...
+            azure_deployment=azure_match[2],
         )
 
-        if match:
-            return AzureOpenAIEndpoint(
-                azure_endpoint=match[1],
-                azure_deployment=match[2],
-            )
+    if openai_match := re.search(f"(.+?)/{name}", endpoint):
+        return OpenAIEndpoint(base_url=openai_match[1])
```
adubovik, line 66: Python returns None by default when there are no return statements and thus no return value.

However, here it's a slightly different situation - we do have have a return value which may be `None`.

So it would be great to see the final `return None`, even though it works as is just fine.

aidial_adapter_openai/utils/parsers.py
```diff
@@ -52,32 +52,45 @@ def get_client(self, params: OpenAIParams) -> AsyncOpenAI:
...
 
+    def parse(self, endpoint: str) -> AzureOpenAIEndpoint | OpenAIEndpoint:
+        if result := _parse_endpoint(self.name, endpoint):
+            return result
         raise DialException(
             "Invalid upstream endpoint format", 400, "invalid_request_error"
         )
 
 
+class LegacyCompletionsParser(BaseModel):
```
adubovik: It doesn't really bring anything of value that we point out here that `/completions` endpoint is **legacy**.

Moreover, it's only legacy in **OpenAI**, but many other models _(e.g. one from HuggingFace)_ do support completions mode and this might be the only mode they support. And thus there is nothing legacy about it.

adubovik: I suggest find-and-replace "legacy" -> "" throughout the project.

roman-romanov-o: Fixed, removed "legacy*"

aidial_adapter_openai/utils/parsers.py
roman-romanov-o: Fixed

aidial_adapter_openai/app.py

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-124.nightly-test.deltixhub.io
>E2E tests status: success

adubovik (approved):

PR #131, opened 2024-07-04
fix: failing with error for incorrect GPT4V request

### LINKED ISSUE #108: UX: throw a proper user error in case when attachments aren't correct
Currently GPT4 Vision and GPT4-O models return `Error` stage and an **empty** `content` field when one of the file attachments isn't of one of the supported types. This confuses the user into assuming that nothing has gone wrong and one could continue sending new messages. 

However, it's a genuine error and must be reported as such with 400 error code.

adubovik:
Resolves #108


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-131.nightly-test.deltixhub.io
>E2E tests status: failed

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-131.nightly-test.deltixhub.io
>E2E tests status: failed

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-131.nightly-test.deltixhub.io
>E2E tests status: failed

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-131.nightly-test.deltixhub.io
>E2E tests status: failed

aidial_adapter_openai/gpt4_multi_modal/download.py
```diff
@@ -23,30 +24,55 @@
 
 def guess_attachment_type(attachment: dict) -> Optional[str]:
     type = attachment.get("type")
-    if type is None:
-        return None
 
-    if "octet-stream" in type:
+    if type is None or "octet-stream" in type:
```
roman-romanov-o, line 28: Could you please explain reasoning of such change?

I understand, that there is a possibility of guessing file type correctly for such cases

```json
{
"type": null,
"url": "http://test.com/image.png"
}
```

But I think it's better to enforce user to set types for attachments, as we did before, don't you think?

aidial_adapter_openai/gpt4_multi_modal/download.py
```diff
@@ -23,30 +24,55 @@
...
         # It's an arbitrary binary file. Trying to guess the type from the URL.
         url = attachment.get("url")
         if url is not None:
             url_type = mimetypes.guess_type(url)[0]
             if url_type is not None:
                 return url_type
-        return None
```
roman-romanov-o, line 36: I think we arranged on one of previous PRs, that even if statement will still fall into returning None anyway, it's better to make this `return None` explicit

aidial_adapter_openai/gpt4_multi_modal/download.py
```diff
@@ -23,30 +24,55 @@
...
+        return ImageFail(name=name, message=message)
+
     try:
         type = guess_attachment_type(attachment)
         if type is None:
-            return "Can't derive media type of the attachment"
+            return fail("can't derive media type of the attachment")
         elif type not in SUPPORTED_IMAGE_TYPES:
-            return f"The attachment isn't one of the supported types: {type}"
+            return fail("the attachment is not one of the supported types")
```
roman-romanov-o, line 75: Why do you remove this part `types: {type}"`?
It's additional info, that could be helpful

aidial_adapter_openai/gpt4_multi_modal/download.py
```diff
@@ -56,12 +82,11 @@ async def download_image(
...
             image_url = ImageDataURL.from_data_url(attachment_link)
             if image_url is not None:
-                if image_url.type in SUPPORTED_IMAGE_TYPES:
-                    return image_url
-                else:
-                    return (
-                        "The image attachment isn't one of the supported types"
+                if image_url.type not in SUPPORTED_IMAGE_TYPES:
+                    return fail(
+                        "the attachment is not one of the supported types"
```
roman-romanov-o, line 87:
```python
return fail(f"the attachment is not one of the supported types: {image_url.type}")
```

aidial_adapter_openai/gpt4_multi_modal/download.py
```diff
@@ -131,27 +154,23 @@ async def transform_message(
...
+            errors.update(result)
         else:
             new_message, tokens = result
             new_messages.append(new_message)
             image_tokens += tokens
 
     if errors:
-        msg = "Some of the image attachments failed to download:"
-        for i, error in errors.items():
-            msg += f"\n- {format_ordinal(i)} message from end:"
```
roman-romanov-o, line 152: minor:

I think that's pretty handful feature, especially for chat users, who will see this as an error stage, to point exact message, where problematic attachment was

aidial_adapter_openai/utils/storage.py
```diff
@@ -51,6 +52,15 @@ async def _get_bucket(self, session: aiohttp.ClientSession) -> Bucket:
 
         return self.bucket
 
+    async def _get_user_bucket(self, session: aiohttp.ClientSession) -> str:
+        bucket = await self._get_bucket(session)
+        appdata = bucket.get("appdata")
```
roman-romanov-o, line 57: But why we don't use `bucket` field from the [API](https://epam-rail.com/dial_api#tag/Files/paths/~1v1~1bucket/get) response?

Due to schema, it's required, on contrary to `appdata`

aidial_adapter_openai/gpt4_multi_modal/download.py
adubovik: What we did before was too restrictive.

`type` is an optional field, meaning we have to do our best to figure out the type from `url` or `data` when it's absent, otherwise, it didn't make any sense to make it optional in the first place.

Consider it's being `None` as the same as it's being `application/octet-stream` - it's declaration of the user's ignorance about the type of the attachment.

aidial_adapter_openai/gpt4_multi_modal/download.py
adubovik: I would have agreed if the method ended right here, however, it goes on up to the explicit return: `return type`.

aidial_adapter_openai/gpt4_multi_modal/download.py
adubovik: It's debatable: first of all, this message is a **user-facing** error message, meaning that the Chat user will see it in a red error box.

Secondly, `type` is a **MIME type**. So arguably this is not a very useful information the Chat user, since it's very likely one doesn't have any notion of MIME types. One would rather familiar with file extensions.

And finally, we print the usage message enumerating the supported file extensions and that should be enough of a help for the user to figure out why the attachment isn't correct.

aidial_adapter_openai/gpt4_multi_modal/download.py
adubovik: https://github.com/epam/ai-dial-adapter-openai/pull/131#discussion_r1668287679

aidial_adapter_openai/gpt4_multi_modal/download.py
adubovik: The problematic attachment in most of the cases will be attached to the **last message**, because we do not allow to continue conversation with an incorrect attachment. So if something went wrong, it's more than likely the problem is the last message.

The exception is when one has changed one model to another. But that should be rare.

aidial_adapter_openai/utils/storage.py
adubovik: Because it's two different things:

1. `bucket` - is a bucket of the **application**. It identifies its file space.
2. `appdata` - points to the **user**'s file space, in particular, it points to a folder which the application has the rights to read from and write to.

E.g.
```json
{
    "bucket": "<application_bucket>",
    "appdata": "<john_smith_bucket>/appdata/<application_deployment_id>"
}
```

When the application recieves DIAL URL to a file, it points to the **user**'s file space:

```json 
{
    "url": "files/<john_smith_bucket>/foo/bar/document.pdf"
}
```

So we need to use `appdata` to extract the user's bucket and remove it from the attachment url to make a readable filename without any redundant information.

roman-romanov-o (approved):

PR #138, opened 2024-08-07
feat: avoid computing token usage if upstream model has reported it already

adubovik:
The fix allows to use upstream models which are able to report usages in streaming mode _(e.g. the one from bedrock/vertexai adapter)_ without token counting issues.

### DIFF
--- a/aidial_adapter_openai/gpt.py
+++ b/aidial_adapter_openai/gpt.py
@@ -56,15 +56,16 @@ async def gpt_chat_completion(
 
     if isinstance(response, AsyncStream):
 
-        prompt_tokens = tokenizer.calculate_prompt_tokens(data["messages"])
         return StreamingResponse(
             to_openai_sse_stream(
                 generate_stream(
-                    prompt_tokens,
-                    map_stream(chunk_to_dict, response),
-                    tokenizer,
-                    deployment_id,
-                    discarded_messages,
+                    get_prompt_tokens=lambda: tokenizer.calculate_prompt_tokens(
+                        data["messages"]
+                    ),
+                    tokenize=tokenizer.calculate_tokens,
+                    deployment=deployment_id,
+                    discarded_messages=discarded_messages,
+                    stream=map_stream(chunk_to_dict, response),
                 ),
             ),
             media_type="text/event-stream",
--- a/aidial_adapter_openai/utils/streaming.py
+++ b/aidial_adapter_openai/utils/streaming.py
@@ -63,7 +63,11 @@ async def generate_stream(
     stream_finished = False
     total_content = ""
 
-    def finalize_last_chunk(chunk: dict) -> None:
+    def finalize_finish_chunk(chunk: dict) -> None:
+        """
+        Adding additional information to a chunk that has non-null finish_reason field.
+        """
+
         if not chunk.get("usage"):
             completion_tokens = tokenize(total_content)
             prompt_tokens = get_prompt_tokens()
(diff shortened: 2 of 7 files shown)


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-138.nightly-test.deltixhub.io
>E2E tests status: failed

aidial_adapter_openai/gpt.py
```diff
@@ -56,15 +56,16 @@ async def gpt_chat_completion(
...
-        prompt_tokens = tokenizer.calculate_prompt_tokens(data["messages"])
         return StreamingResponse(
             to_openai_sse_stream(
                 generate_stream(
-                    prompt_tokens,
-                    map_stream(chunk_to_dict, response),
-                    tokenizer,
-                    deployment_id,
-                    discarded_messages,
+                    get_prompt_tokens=lambda: tokenizer.calculate_prompt_tokens(
```
roman-romanov-o, line 62: What's the point of wrapping this in lambda, why don't pass the value itself?

aidial_adapter_openai/utils/streaming.py
```diff
@@ -51,42 +51,45 @@ def build_chunk(
...
     deployment: str,
     discarded_messages: Optional[list[int]],
+    stream: AsyncIterator[dict],
 ) -> AsyncIterator[dict]:
+
     last_chunk, temp_chunk = None, None
     stream_finished = False
+    total_content = ""
+
+    def finalize_last_chunk(chunk: dict) -> None:
```
roman-romanov-o: Let's avoid any functions with hidden changes of the object
When I see this
```python
async for chunk in stream:
    finalize_last_chunk(chunk)
```
I don't know from first glance, that chunk was modified, on contrary to


```python
async for chunk in stream:
    chunk = finalize_last_chunk(chunk)
```

aidial_adapter_openai/utils/streaming.py
```diff
@@ -106,39 +109,26 @@ async def generate_stream(
...
+                created=str(int(time())),
+                message={},
+                finish_reason=None,
             )
+        else:
+            logger.warning("Didn't receive chunk with the finish reason")
+
+        finalize_last_chunk(last_chunk)
+
+        last_chunk["choices"][0]["delta"] = {}
```
roman-romanov-o: Last chunk could be some content, so it would be better to make
```python
last_chunk["choices"][0]["delta"] = last_chunk["choices"][0].get('delta') or {}
```

roman-romanov-o: Also, is there any case, when choices would be just empty array?

aidial_adapter_openai/gpt.py
adubovik: To avoid running tokenization if the upstream model has done it on its own.

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-138.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/utils/streaming.py
adubovik: The code was a bit confusing indeed. I've refactored it.

Empty array of choices does appear in cases when the model has nothing to report in the content of choices, but reports some auxilary information like usage.

Let's handle the case with multipe choices in a separate PR: https://github.com/epam/ai-dial-adapter-openai/issues/140

roman-romanov-o (approved):

PR #150, opened 2024-09-09
feat: add message truncation for multi modal

### LINKED ISSUE #149: Support prompt truncation for GPT4-o models
Note that it's already [supported](https://github.com/epam/ai-dial-adapter-openai/blob/2dd46673e7628bc627c0e78be0a22d32700cb631/aidial_adapter_openai/gpt.py#L41-L43) for the regular GPT-4 models.

roman-romanov-o:
resolves https://github.com/epam/ai-dial-adapter-openai/issues/149


roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-150.nightly-test.deltixhub.io
>E2E tests status: success

README.md
```diff
@@ -64,6 +64,7 @@ Copy `.env.example` to `.env` and customize it for your environment:
 |DALLE3_DEPLOYMENTS|``|Comma-separated list of deployments that support DALL-E 3 API. Example: `dall-e-3,dalle3,dall-e`|
 |GPT4_VISION_DEPLOYMENTS|``|Comma-separated list of deployments that support GPT-4V API. Example: `gpt-4-vision-preview,gpt-4-vision`|
 |GPT4_VISION_MAX_TOKENS|1024|Default value of `max_tokens` parameter for GPT-4V when it wasn't provided in the request|
+|GPT4O_MAX_TOKENS|4096|Default value of `max_tokens` parameter for GPT-4o when it wasn't provided in the request|
```
adubovik: We use to provide default max_tokens value in one of the cases:
1. The default max_tokens of the model itself is set to be very low (it was the case for GPT4-V, its default was only 16 tokens)
2. The model does allow missing max_tokens argument, so we have to provide one with every request.

Does GPT4-o fall into one of the cases?

Also, keep in mind that the need for max_tokens defaults has lessen significantly since DIAL Core supported `defaults` for chat completion request overrides.

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -41,6 +48,7 @@
 # The built-in default max_tokens is 16 tokens,
 # which is too small for most image-to-text use cases.
 GPT4V_DEFAULT_MAX_TOKENS = int(os.getenv("GPT4_VISION_MAX_TOKENS", "1024"))
+GPT4O_DEFAULT_MAX_TOKEN = int(os.getenv("GPT4O_MAX_TOKENS", "4096"))
```
adubovik: `GPT4O_DEFAULT_MAX_TOKENS`

aidial_adapter_openai/gpt4_multi_modal/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+    custom_content = message.get("custom_content", {})
+    attachments = custom_content.get("attachments", [])
+
+    message = {k: v for k, v in message.items() if k != "custom_content"}
+
+    if len(attachments) == 0:
+        return MessageTransformResult(
+            message=message,
+            text_tokens=tokenizer.calculate_tokens_per_message(message),
+        )
```
adubovik: This shortcut seems redundant. Nothing really would change performace-wise, semantic-wise if you remove it. The best code is one that is never written.

aidial_adapter_openai/gpt4_multi_modal/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+        tokens, detail = tokenize_image(image_url, "auto")
+        image_tokens.append(tokens)
+        image_messages.append(create_image_message(image_url, detail))
+
+    logger.debug(f"image tokens: {image_tokens}")
+
+    return MessageTransformResult(
+        image_tokens=sum(image_tokens),
+        text_tokens=tokenizer.calculate_tokens_per_message(
+            create_text_message(content)
```
adubovik: `create_text_message` returns a GPT4-o **sub-message**.

But `calculate_tokens_per_message` expectes GPT4 **message**.

aidial_adapter_openai/gpt4_multi_modal/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+    image_tokens: int = 0
+    errors: Optional[List[ImageFail]] = None
+    message: dict
+
+    @property
+    def total_tokens(self) -> int:
+        return self.text_tokens + self.image_tokens
+
+
+class TransformMessagesResult(BaseModel):
```
adubovik: It's two types sqeezed into one.

Why not `List[MessageTransformResult] | str`?

It's 14 lines shorter and provides better type-level guarantees - it's either this or that.

aidial_adapter_openai/gpt4_multi_modal/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+            values.get("error_message") is None
+            and values.get("transformations") is None
+        ):
+            raise ValueError(
+                "Either 'error_message' or 'transformations' must be provided."
+            )
+        return values
+
+
+class TruncateTransformedMessagesResult(BaseModel):
```
adubovik: Please move to a dedicated `truncate` module.

It's not used in the current module. 
This module is about conversion from GPT4 request to GPT-4o request.

aidial_adapter_openai/gpt4_multi_modal/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+    create_image_message,
+    create_text_message,
+)
+from aidial_adapter_openai.utils.image_data_url import ImageDataURL
+from aidial_adapter_openai.utils.log_config import logger
+from aidial_adapter_openai.utils.storage import FileStorage
+from aidial_adapter_openai.utils.tokens import Tokenizer
+
+
+class MessageTransformResult(BaseModel):
```
adubovik: Why not split into two types:

`MessageTransformResult = MessageTransformError | MessageTransformSuccess`

aidial_adapter_openai/gpt4_multi_modal/transformation.py
adubovik: There are two ways to name thing:
1. Reflect in its name how it was produced - to answer the question _What is it the result of_?
2. Reflect what is it semantically - _What does it represent? How could it be used downstream?_

First hightlights past of an object, second - its future.

The second is more commonly used, because we rathely care about the past of objects.

So maybe `GPT4OMessages` would be a better name.

aidial_adapter_openai/gpt4_multi_modal/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+        logger.error(f"download errors: {errors}")
+        return MessageTransformResult(message=message, errors=errors)
+
+    image_urls: List[ImageDataURL] = cast(List[ImageDataURL], download_results)
+
+    image_tokens: List[int] = []
+    image_messages: List[dict] = []
+
+    for image_url in image_urls:
+        tokens, detail = tokenize_image(image_url, "auto")
```
adubovik: Optional: factoring out this token-counting logic into a dedicated `tokenizer` module would likely result in a clearer codebase.

aidial_adapter_openai/gpt4_multi_modal/truncate.py
```diff
@@ -0,0 +1,63 @@
...
+    TruncatePromptSystemError,
+)
+
+from aidial_adapter_openai.gpt4_multi_modal.transformation import (
+    MessageTransformResult,
+    TruncateTransformedMessagesResult,
+)
+
+
+def truncate_transformed_messages(
```
adubovik: Keep it simple `def truncate_prompt`

aidial_adapter_openai/gpt4_multi_modal/truncate.py
```diff
@@ -0,0 +1,63 @@
...
+    transformations: List[MessageTransformResult], max_prompt_tokens: int
```
adubovik: It's basically a copy-paste of `aidial_adapter_openai.utils.tokens:truncate_prompt`.
Why not reuse with a sufficient generalization?

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -180,26 +188,31 @@ async def chat_completion(
...
 
-    new_messages, prompt_image_tokens = result
+    max_tokens = request.get("max_tokens") or default_max_tokens
 
-    prompt_text_tokens = tokenizer.calculate_prompt_tokens(messages)
-    estimated_prompt_tokens = prompt_text_tokens + prompt_image_tokens
+    if result.transformations is None:
+        raise InternalServerError("Transformations cannot be null")
+    truncation_result = truncate_transformed_messages(
+        result.transformations, max_tokens
```
adubovik: `request.max_tokens` and `request.max_prompt_tokens` are two different parameters.
Here you need to pass `request.max_prompt_tokens`.

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -180,26 +188,31 @@ async def chat_completion(
...
+    )
 
     request = {
         **request,
-        "max_tokens": request.get("max_tokens") or default_max_tokens,
-        "messages": new_messages,
+        "max_tokens": max_tokens,
+        "messages": truncation_result.messages,
     }
```
adubovik, line 241: Destructure to reduce verbosity below.

`estimated_prompt_tokens, discarded_messages = truncation_result.overall_token, truncation_result.discarded_messages`

aidial_adapter_openai/gpt4_multi_modal/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ValueError(
+                "Either 'error_message' or 'transformations' must be provided."
+            )
+        return values
+
+
+class TruncateTransformedMessagesResult(BaseModel):
+    messages: List[dict]
+    discarded_messages: List[int]
+    overall_token: int
```
adubovik: `overall_token => prompt_tokens` or even `tokens`.

`overall` doesn't add much to the meaning of the field.

README.md
roman-romanov-o: Fixed, removed

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: Fixed, removed

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: Fixed, removed

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: Fixed, pass whole message to tokenizer

Modified tokenizer to work with sub-messages

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: I think that returning error as "str" is not verbose enough for first-glance


I've changed result into two dataclasses


```python
class MultiModalMessage(BaseModel):
    tokens: int
    message: dict


class TransformationError(BaseModel):
    image_fails: List[ImageFail]

```

and result of function would be
```python
-> List[MultiModalMessage] | TransformationError
```

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: Fixed

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: https://github.com/epam/ai-dial-adapter-openai/pull/150#discussion_r1763345034

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: Came up with `MultiModalMessage` naming, since we can use this truncation not only for `GPT40`, but for `GPT4Vision` also

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: Done, moved to `image_tokenizer` module

aidial_adapter_openai/gpt4_multi_modal/truncate.py
roman-romanov-o: Named it as `mutlimodal_truncate_prompt`, since now we have generalized `truncate_prompt`

aidial_adapter_openai/gpt4_multi_modal/truncate.py
roman-romanov-o: Good catch
Generalized this in one function

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: Fixed

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: Not actual anymore, function has changed drastically

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: Fixed

tests/test_discard_messages.py
```diff
@@ -3,16 +3,19 @@
...
+    PlainMessages,
+    MaxPromptTokens,
+    Tuple[PlainMessages, DiscardedMessages],
+]
 
 normal_cases: List[TestCase] = [
-    (
-        [],
-        0,
-        ([], []),
```
roman-romanov-o, line 14: I had to remove this tests case, because it actually doesn't cover logic of tuncations
Correct me, if I'm wrong

Any message would actually have initial amount of tokens assigned (3 tokens)

So we can't actually say that if you set `max_prompt_tokens = 0`, you will actually use 0 tokens

The only reason this worked before, is becasue there was explicit if-statement, that covered this case

```python
 if n == 0:
        return messages, []  # will be rejected by the upstream
```

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-150.nightly-test.deltixhub.io
>E2E tests status: failed

aidial_adapter_openai/utils/tokens.py
```diff
@@ -43,62 +44,23 @@ def calculate_tokens_per_message(self, message: dict) -> int:
 
         tokens = tokens_per_message
         for key, value in message.items():
-            if isinstance(value, str):
-                tokens += self.calculate_tokens(value)
             if key == "name":
                 tokens += tokens_per_name
-
+            elif isinstance(value, str):
```
adubovik: Maybe this structure? It looks much more straight-forward:

```python
if name == "name": ...
elif name == "content":
    if isinstance(value, str): ....
    elif isinstance(value, list): ....
```

aidial_adapter_openai/utils/tokens.py
```diff
@@ -2,20 +2,24 @@
...
+    """
+    Tokenizer for message.
+    Calculates only textual tokens, not image tokens.
+    """
+
     model: str
     encoding: Encoding
 
+    # Tokens that are added to all message tokens, not matter what
+    PROMPT_TOKENS = 3
```
adubovik: It's not prompt tokens per se, it's rather `_TOKENS_PER_REQUEST` - extra for a request. Similar to what we have for `tokens_per_message` and `tokens_per_name` - extra tokens for these entities.

aidial_adapter_openai/utils/tokens.py
```diff
@@ -2,20 +2,24 @@
 Implemented based on the official recipe: https://cookbook.openai.com/examples/how_to_count_tokens_with_tiktoken
```
adubovik: Maybe worth to rename to `tokenizer.py` since it's the only thing here now.

aidial_adapter_openai/utils/prompt_truncate.py
```diff
@@ -0,0 +1,65 @@
+from typing import Callable, List, Set, Tuple, TypeVar
```
adubovik: Why the inversion `prompt_truncate`? It would have made sense for `prompt_truncation`, but when you have a verb, the more common order of words is `truncate_prompt` - same as the method in the module and an endpoint in DIAL API.

aidial_adapter_openai/utils/prompt_truncate.py
```diff
@@ -0,0 +1,65 @@
...
+_T = TypeVar("_T")
+
+DiscardedMessages = List[int]
+UsedTokens = int
+
+
+def truncate_prompt(
+    message_holders: List[_T],
+    message_tokens_getter: Callable[[_T], int],
+    is_system_message_getter: Callable[[_T], bool],
```
adubovik: Why `_getter` suffix? Usually predicates like this one are called `is_SOMETHING`: `is_system_message`. Now it's clear from the name that it's a method, not an object.

Likewise for `message_tokens_getter` - if you call a variable using a verb you kill two birds with one stone: (1) it becomes clear it's a method (2) you don't have to spend character on the getter suffix: `tokenize_message`.

aidial_adapter_openai/utils/prompt_truncate.py
```diff
@@ -0,0 +1,65 @@
...
+)
+
+_T = TypeVar("_T")
+
+DiscardedMessages = List[int]
+UsedTokens = int
+
+
+def truncate_prompt(
+    message_holders: List[_T],
```
adubovik: What value what does the `_holders` suffix provide? `messages` is as clear as it could be.

aidial_adapter_openai/utils/prompt_truncate.py
```diff
@@ -0,0 +1,65 @@
...
+        raise TruncatePromptSystemError(max_prompt_tokens, prompt_tokens)
+
+    # Then non-system messages in the reverse order
+    for idx, message_holder in reversed(list(enumerate(message_holders))):
+        if is_system_message_getter(message_holder):
+            continue
+        message_tokens = message_tokens_getter(message_holder)
+
+        if max_prompt_tokens < prompt_tokens + message_tokens:
+            if len(
```
adubovik: `system_messages_count != len(message_holders)` is always True at this point.

aidial_adapter_openai/utils/prompt_truncate.py
```diff
@@ -0,0 +1,65 @@
...
+DiscardedMessages = List[int]
+UsedTokens = int
+
+
+def truncate_prompt(
+    message_holders: List[_T],
+    message_tokens_getter: Callable[[_T], int],
+    is_system_message_getter: Callable[[_T], bool],
+    max_prompt_tokens: int,
+    initial_prompt_tokens: int = 3,
```
adubovik: The default here is strange - why is 3 the best default value for this **generic algorithm**? 3 comes from OpenAI tokenizer, I would have passed it explicitly.

aidial_adapter_openai/gpt4_multi_modal/transformation.py
adubovik: I meant the tokenization algo for the whole request, not only for images.
Here transformation both (1) downloads images and created GPT4V compatible message (2) tokenizes the request.

I suggest to separate the two for modularity and ease of testing.

As you have a `Tokenizer` for text OpenAI models, you could introduce `VisionTokenizer` for GPT4V specifically which is going to handle images gracefully and reuse the regular `Tokenizer` where it's needed.

aidial_adapter_openai/gpt.py
```diff
@@ -14,7 +21,20 @@
...
-from aidial_adapter_openai.utils.tokens import Tokenizer, truncate_prompt
+from aidial_adapter_openai.utils.tokens import Tokenizer
+
+
+def plain_message_truncate(
+    messages: List[dict], max_prompt_tokens: int, tokenizer: Tokenizer
+) -> Tuple[List[dict], DiscardedMessages, UsedTokens]:
+    return truncate_prompt(
+        message_holders=messages,
+        message_tokens_getter=lambda message: tokenizer.calculate_tokens_per_message(
```
adubovik: `message_tokens_getter=tokenizer.calculate_tokens_per_message`

aidial_adapter_openai/gpt.py
```diff
@@ -14,7 +21,20 @@
     generate_stream,
     map_stream,
 )
-from aidial_adapter_openai.utils.tokens import Tokenizer, truncate_prompt
+from aidial_adapter_openai.utils.tokens import Tokenizer
+
+
+def plain_message_truncate(
```
adubovik: Maybe `plain_text_truncate_prompt` to elaborate on what `plain` actually mean? Compare with `multimodal_truncate_prompt` in another module.

aidial_adapter_openai/gpt.py
```diff
@@ -26,6 +46,7 @@ async def gpt_chat_completion(
     tokenizer: Tokenizer,
 ):
     discarded_messages = None
+    used_prompt_tokens = None
```
adubovik: Maybe simply `prompt_tokens`.
When you add an adjective to a name of a variable, it implies that there are other kinds of this thing differing in this aspect.
That is, `used_prompt_tokens` implies the existence of unused or eastimated prompt tokens.

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -41,6 +48,7 @@
 # The built-in default max_tokens is 16 tokens,
 # which is too small for most image-to-text use cases.
 GPT4V_DEFAULT_MAX_TOKENS = int(os.getenv("GPT4_VISION_MAX_TOKENS", "1024"))
+# GPT4O_DEFAULT_MAX_TOKEN = int(os.getenv("GPT4O_MAX_TOKENS", "4096"))
```
adubovik: Leftover comment

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -180,26 +201,45 @@ async def chat_completion(
...
-    new_messages, prompt_image_tokens = result
-
-    prompt_text_tokens = tokenizer.calculate_prompt_tokens(messages)
-    estimated_prompt_tokens = prompt_text_tokens + prompt_image_tokens
+        return create_response_from_chunk(
+            chunk, transformation_result, is_stream
+        )
+    max_prompt_tokens = None
+    if "max_prompt_tokens" in request:
+        max_prompt_tokens = request.pop("max_prompt_tokens")
```
adubovik: Could be simplified to `max_prompt_tokens = request.pop("max_prompt_tokens", None)`

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -180,26 +201,45 @@ async def chat_completion(
...
+
+    truncation_result = transformation_result
+    discarded_messages = None
+    estimated_propmpt_tokens = tokenizer.calculate_overall_prompt_tokens(
```
adubovik: `propmpt` typo

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -180,26 +201,45 @@ async def chat_completion(
...
-    estimated_prompt_tokens = prompt_text_tokens + prompt_image_tokens
+        return create_response_from_chunk(
+            chunk, transformation_result, is_stream
+        )
+    max_prompt_tokens = None
+    if "max_prompt_tokens" in request:
+        max_prompt_tokens = request.pop("max_prompt_tokens")
+
+    truncation_result = transformation_result
+    discarded_messages = None
```
adubovik, line 214: Move the line as close to the second assignment as possible to easy reading.

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -111,6 +119,20 @@ async def predict_non_stream(
             return await response.json()
 
 
+def multimodal_truncate_prompt(
+    transformations: List[MultiModalMessage],
```
adubovik: `transformations` here doesn't make sense. 
We don't really care about the past of this object, we care about what it actually [represents](https://github.com/epam/ai-dial-adapter-openai/pull/150#discussion_r1752113762) now. I suggest simply `messages`.

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
```diff
@@ -249,10 +289,15 @@ def debug_print(chunk: T) -> T:
         content = response["choices"][0]["message"].get("content") or ""
         usage = response["usage"]
 
+        if discarded_messages:
+            response |= {
+                "statistics": {"discarded_messages": discarded_messages}
+            }
+
```
adubovik, line 296: This condition isn't required, since `generate_stream` took care of it already.

tests/test_discard_messages.py
```diff
@@ -3,16 +3,19 @@
 import pytest
 from aidial_sdk.exceptions import HTTPException as DialException
 
-from aidial_adapter_openai.utils.tokens import Tokenizer, truncate_prompt
+from aidial_adapter_openai.gpt import plain_message_truncate
+from aidial_adapter_openai.utils.prompt_truncate import DiscardedMessages
+from aidial_adapter_openai.utils.tokens import Tokenizer
 
-TestCase = Tuple[List[dict], int, Tuple[List[dict], List[int]] | str]
+PlainMessages = List[dict]
```
adubovik: `PlainTextMessages` maybe.

aidial_adapter_openai/utils/tokens.py
```diff
@@ -43,62 +44,23 @@ def calculate_tokens_per_message(self, message: dict) -> int:
...
-
+            elif isinstance(value, str):
+                tokens += self.calculate_tokens(value)
+            if key == "content" and isinstance(value, list):
+                for submessage in value:
+                    if submessage["type"] == "text":
+                        tokens += self.calculate_tokens(submessage["text"])
         return tokens
 
+    def calculate_overall_prompt_tokens(self, message_tokens: int):
```
adubovik: `overall` here actually means **the whole request**. So let's spell it out explicitly: `calculate_request_prompt_tokens`.

Also `message_tokens` is rather `messages_tokens`, because it supposed to be a sum of tokens for **all messages** in the request.

aidial_adapter_openai/utils/prompt_truncate.py
roman-romanov-o: Because this way user actually understand, that we pass not the value itself, but function, that would get us the value

```python
test_func(is_system_message=test_value) 
```
Here I can easily assume, that we pass boolean values

---

```python
test_func(is_system_message_getter=test_value)
```
Here I can explicitly see, that this argument is not primitive type, but a getter

aidial_adapter_openai/gpt4_multi_modal/transformation.py
roman-romanov-o: Moved all this logic into `utils/tokenizer.py`

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-150.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/utils/tokens.py
roman-romanov-o: Fixed

tests/test_discard_messages.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt4_multi_modal/chat_completion.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt.py
roman-romanov-o: fixed

aidial_adapter_openai/gpt.py
roman-romanov-o: fixed

aidial_adapter_openai/utils/prompt_truncate.py
roman-romanov-o: fixed

aidial_adapter_openai/utils/prompt_truncate.py
roman-romanov-o: Because it's not exactly the *message*. For multi-modal have object, that bounds message with data about pictures.

So I want to explicitly separate `message`, that we operate frequently, and `message_holder` - object, that *contains message*, but not bound by that

That's why argument for this function are also `is_system_message_getter` - because we can add custom logic to extract system message from this object

aidial_adapter_openai/utils/prompt_truncate.py
roman-romanov-o: Fixed

aidial_adapter_openai/utils/tokens.py
roman-romanov-o: Fixed

aidial_adapter_openai/utils/tokens.py
roman-romanov-o: Renamed to `TOKENS_PER_REQUEST`, without `_T...` prefix, since I need this property outside of class

aidial_adapter_openai/utils/tokens.py
roman-romanov-o: Why do we need to introduce this extra nesting?

Content with string in value will be processed by 
```python
elif isinstance(value, str):
```
as any other key in message

with this `if` I want to cover one case, and one case only - when content has custom content in it

aidial_adapter_openai/app.py
```diff
@@ -44,7 +44,10 @@
 )
 from aidial_adapter_openai.utils.reflection import call_with_extra_body
 from aidial_adapter_openai.utils.storage import create_file_storage
-from aidial_adapter_openai.utils.tokens import Tokenizer
+from aidial_adapter_openai.utils.tokenizer import (
+    MutliModalTokenizer,
```
adubovik: `Mutli => Multi`

aidial_adapter_openai/utils/tokenizer.py
```diff
@@ -0,0 +1,125 @@
...
+from abc import abstractmethod
+from typing import Generic, List, TypeVar
+
+from aidial_sdk.exceptions import InternalServerError
+from tiktoken import Encoding, encoding_for_model
+
+from aidial_adapter_openai.utils.image_tokenizer import tokenize_image_by_size
+from aidial_adapter_openai.utils.multi_modal_message import MultiModalMessage
+
+TokenizeInput = TypeVar("TokenizeInput")
```
adubovik: The tokenizer input is a list of these things, not this thing itself.
Why not `MessageType`?

aidial_adapter_openai/utils/tokenizer.py
```diff
@@ -0,0 +1,125 @@
...
+        for key, value in message.items():
+            if key == "name":
+                tokens += self.tokens_per_name
+            elif key == "content" and isinstance(value, list):
+                for submessage in value:
+                    # Process text part of the message
+                    if submessage["type"] == "text":
+                        tokens += self.calculate_text_tokens(submessage["text"])
+            elif isinstance(value, str):
+                tokens += self.calculate_text_tokens(value)
```
adubovik, line 116: What if a submessage has `type == image`? This method **won't** throw an error.

This bug happens partially because of such a traversal of the structure, which doesn't really follow the type of this structure.

For such a simple map-reduce aggregation you typically visiting every element of a structure **once** in a top-down direction following the types - that is handling all possible cases of a value at the given node there and then.

But here the `key == "content"` is checked twice, and certain cases aren't handled **inplace** _(like type=image)_ and instead the error handling is delegated to the catch-all else guard. This leads to an error-prone code, hard to maintain and read code.

How about this:

```
for key, value in message_holder.items():
    if key == "name": collect
    elif key == "content":
        if isinstance(value, list):
            for submessage in value:
                if submessage["type"] == "text": collect
                else: fail
        elif isinstance(value, str): collect
        else: fail
```

aidial_adapter_openai/utils/tokenizer.py
```diff
@@ -0,0 +1,125 @@
...
+                for submessage in value:
+                    if submessage["type"] == "text":
+                        tokens += self.calculate_text_tokens(submessage["text"])
+            elif isinstance(value, str):
+                tokens += self.calculate_text_tokens(value)
+            elif key == "content" and not isinstance(value, str):
+                raise InternalServerError(
+                    f"Unexpected type of content in message: {value!r}"
+                    f"Use MultiModalTokenizer for messages with images"
+                )
```
adubovik: Why splitting handling of the `key == "content"` case into two branches? It's very confusing. The meaning of the second `content` branch depends very heavily on what has been covered by the branch above. This increases mental load significantly. 

Why not traverse every single element of the structure once and deal with it there and then, without smearing the logic across multiple lines:

```
for key, value in message_holder.items():
    if key == "name": collect
    elif key == "content":
        if isinstance(value, list):
            for submessage in value:
                if submessage["type"] == "text": collect
                else: fail
        elif isinstance(value, str): collect
        else: fail
```

This decreases the coupling of the code and very easy to understand, since it's mechanically derived from the definition of the ChatCompletionChunk.

aidial_adapter_openai/utils/tokenizer.py
```diff
@@ -0,0 +1,125 @@
...
+        tokens = self.tokens_per_message
+        for key, value in message_holder.items():
+            if key == "name":
+                tokens += self.tokens_per_name
+            elif key == "content" and isinstance(value, list):
+                for submessage in value:
+                    if submessage["type"] == "text":
+                        tokens += self.calculate_text_tokens(submessage["text"])
+            elif isinstance(value, str):
+                tokens += self.calculate_text_tokens(value)
```
adubovik: I would suggest to **ignore** values accosiated with any keys other than "name" and "content".

Otherwise we are going to compute tokens for some irrelevant fields or even worse fail in runtime processing these extra fields:

```json
{
    "name": "user",
    "content": "hello",
    "custom_content": "..."
}
```

Same for the MultiModal case.

aidial_adapter_openai/utils/tokenizer.py
```diff
@@ -0,0 +1,125 @@
...
+        Tokens, that are counter for "name" field in message, if it's present
+        """
+        if self.model == "gpt-3.5-turbo-0301":
+            return -1
+        return 1
+
+    def calculate_request_prompt_tokens(self, messages_tokens: int):
+        """
+        Amount of tokens, that will be counted by API
+        is greater than actual sum of tokens of all messages by PROMPT_TOKENS
```
adubovik: `PROMPT_TOKENS => TOKENS_PER_REQUEST`

Not sure we need this comment for a single line of code with very telling namings...

tests/test_discard_messages.py
adubovik: I agree, but let's set max_prompt_tokens with 3 nonetheless to test the corner case with the empty list:

```
(
        [],
        3,
        ([], [])
)
```

And add a test case with an error to `error_cases`:

```
([], 2, "error_message")
```

aidial_adapter_openai/utils/prompt_truncate.py
adubovik: The types help to resolve the ambiguity in case of `is_system_message`.

What about `tokenize_message`? I don't see any ambiguity here.

aidial_adapter_openai/utils/prompt_truncate.py
adubovik: But the whole point of a **generic** algorithm is that it doesn't care about the particular implementation of the message type `_T`. It could be **anything** - it's not important for the logic of the algorithm. The algorithm couldn't tell a difference between an exact message and a message holder. It's all the same for it.

What's important for the person who reads this code is only the fact that `_T` represents a **message** in a most generic sense. Given this undersanding one could immediately realize in what contexts this algo might be used.

tests/test_discard_messages.py
roman-romanov-o: Fixed

aidial_adapter_openai/app.py

aidial_adapter_openai/utils/tokenizer.py
roman-romanov-o: Fixed

aidial_adapter_openai/utils/tokenizer.py
roman-romanov-o: Fixed

aidial_adapter_openai/utils/tokenizer.py
roman-romanov-o: Not only `name` and `content`, but also `role` is tokenized, for instance

And my concern, is that we can miss that key at some point
--

But actually, when I've changed it to this strict set of keys, I've found out that our unit test was actually falling into case, that you've explained, when message had incorrect format:
```json
{
    "role": "user"
    "message": "..." <---- must be content
}
```

And it was still passing, since "message" was string


So let it be

aidial_adapter_openai/utils/tokenizer.py
roman-romanov-o: I don't see any harm in leaving the comment
It still can add some context to person, that is not familiar with tokenization

aidial_adapter_openai/utils/prompt_truncate.py
roman-romanov-o: Renamed to `messages`

aidial_adapter_openai/utils/prompt_truncate.py
roman-romanov-o: Ok, let the types do explaining

Removed `_getter` postfix

adubovik (dismissed):

adubovik (approved):

PR #151, opened 2024-09-23
fix: reusing usage reported by the upstream OpenAI via include_usage option

### LINKED ISSUE #140: Support token computation for (stream=True AND (include_usage=true AND/OR n>1))
1. `n>1` leads to a stream with **multiple** chunks with `finish_reason!=null`, so we must attach usage only to the last one.
2. The request option

```json
{"stream_options": {"include_usage": true}}
```

leads to a final chunk with an **empty list of choices**:

```json
{
  "choices": [],
  "created": 7777,
  "id": "chatcmpl-8888",
  "model": "gpt-4-turbo-2024-04-09",
  "object": "chat.completion.chunk",
  "system_fingerprint": "fp_e49e4201a9",
  "usage": {
    "completion_tokens": 5,
    "prompt_tokens": 12,
    "total_tokens": 17
  }
}

```

Consider attaching usage not to the last chunk with `finish_reason!=null`, but create a dedicated chunk as it's done when `include_usage=true`.

adubovik:
Resolves https://github.com/epam/ai-dial-adapter-openai/issues/140


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-151.nightly-test.deltixhub.io
>E2E tests status: success

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-151.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/utils/streaming.py
```diff
@@ -1,19 +1,19 @@
...
 from openai import APIError, APIStatusError
 from openai.types.chat.chat_completion_chunk import ChatCompletionChunk
 
 from aidial_adapter_openai.utils.env import get_env_bool
 from aidial_adapter_openai.utils.log_config import logger
+from aidial_adapter_openai.utils.merge_chunks import merge_chunks
 from aidial_adapter_openai.utils.sse_stream import to_openai_sse_stream
 
-fix_streaming_issues_in_new_api_versions = get_env_bool(
+eliminate_empty_choices = get_env_bool(
```
roman-romanov-o: Since it's a constant value from environment, and it's boolean let's use `TO_ELIMINATE_EMPTY_CHOICES`, what do you think?

aidial_adapter_openai/utils/merge_chunks.py
```diff
@@ -0,0 +1,31 @@
...
+
+    assert len(chunks) > 0, "At least one chunk must be provided"
+
+    target = chunks[0]
+
+    def _generator():
+        yield target
+        for source in chunks[1:]:
+            source = source.copy()
+            for key, value in list(source.items()):
```
roman-romanov-o: Why not just 
`for key, value in source.items():`
?

aidial_adapter_openai/utils/merge_chunks.py
```diff
@@ -0,0 +1,31 @@
...
+                if not isinstance(value, (list, dict)) and value is not None:
+                    target[key] = value
+                    del source[key]
+            yield source
+
+    return merge(*list(_generator()))
```
roman-romanov-o: For me it looks kinda scary and over-complicated

We can just use 
```python
from aidial_sdk.utils.merge_chunks import merge_recursive

def merge_chunks(*chunks: _Chunk) -> _Chunk:
    ...
    target = chunks[0]

    for source in chunks[1:]:
        for key, value in list(source.items()):
            if not isinstance(value, (list, dict)) and value is not None:
                target[key] = value
                del source[key]
        target = merge_recursive(target, source, path=[])
    return target
```

What do you think?

aidial_adapter_openai/utils/merge_chunks.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    assert len(chunks) > 0, "At least one chunk must be provided"
+
+    target = chunks[0]
+
+    def _generator():
+        yield target
+        for source in chunks[1:]:
+            source = source.copy()
```
roman-romanov-o: It would be really cool to highlight for user, that we don't want to change actual source

```python
source_to_merge = source.copy()
...

del source_to_merge[key]
```

What do you think?

aidial_adapter_openai/utils/merge_chunks.py
roman-romanov-o: Then this comment from docsrting
```
    Non-atomic fields (e.g. "choice", "usage") are merged following
    the standard recursive merging procedure.
```
actually makes more sense, since he sees `merge_recursive`, and not just `merge` from sdk toolset

aidial_adapter_openai/utils/streaming.py
```diff
@@ -61,84 +61,112 @@ async def generate_stream(
...
 
-    if not stream_finished:
-        if last_chunk is None:
+    if last_chunk is not None and buffer_chunk is not None:
+        last_chunk = merge_chunks(buffer_chunk, last_chunk)
+
+    if discarded_messages is not None:
+        last_chunk = set_discarded_messages(last_chunk, discarded_messages)
+
+    if not found_usage and (not error or completions):
```
roman-romanov-o, line 150: Here, if user received no errors, and no chunks, we would not return to him usage for prompt tokens, why?

If we'll leave just
```python
if not found_usage:
    ...
```
we'll give him prompt tokens count in any case

aidial_adapter_openai/utils/streaming.py
adubovik: Renamed to `ELIMINATE_EMPTY_CHOICES`

aidial_adapter_openai/utils/merge_chunks.py
adubovik: Because I'm removing certain keys in the collection while traversing this very collection.
Python behaviour is undefined in such a case.

aidial_adapter_openai/utils/merge_chunks.py
adubovik: Fixed, but used `merge` instead. No need to reimplement the wheel.

aidial_adapter_openai/utils/merge_chunks.py
adubovik: Fixed via `source = chunk.copy()` to remove ambiguity.

aidial_adapter_openai/utils/streaming.py
adubovik:
```python
if not found_usage and (not error or completions):
    # append estimated usage
```

is equivalent to:

```python
if not found_usage:
    if error and not completions:
        pass
    else:
        # append estimated usage
```

Meaning that the **estimated usage** which hasn't been already reported in the stream isn't reported if:
(1) the server has returned an error AND,
(2) it didn't generate any completion content.

This may happen if the server returns 400, 422 or 500 right away.
It would be wrong to report `usage` _(and `usage.prompt_tokens` in particular)_, because the usage represents actual **payable** tokens. The combination of **(error & no completion generated)** is very unlikely _(though not impossible)_ to incur any charges. So we do not report them.

Compare it with the case of **(error & completion generation)**. It wouldn't be possible for the model to generate something without consuming `prompt_tokens` first AND we already have some generated completion, so we report it as `completion_tokens`.

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-151.nightly-test.deltixhub.io
>E2E tests status: success

roman-romanov-o (approved):

PR #162, opened 2024-10-28
feat: supported multi-modal embeddings from Azure AI Vision service

adubovik:
Introduced a new env var: `AZURE_AI_VISION_DEPLOYMENTS` that contains a comma-separated list of Azure AI Vision embedding deployments. 

The upstreams of such deployments configured in the DIAL Core config are expected to point to Azure AI Vision service: `https://<azure-ai-vision-service-name>.cognitiveservices.azure.com/`

Under the hood the adapter calls `/retrieval:vectorizeImage` and `/retrieval:vectorizeText` [endpoints](https://learn.microsoft.com/en-gb/rest/api/computervision/image-retrieval?view=rest-computervision-v4.0-preview%20(2023-04-01)) of the Azure AI Vision service to compute embeddings for images and texts.

## Example configuration

Replace `{adapter-openai-hostname}` and `{azure-ai-vision-service-name}` with values relevant for our setup.

### DIAL Core config

```json
{
    "models": {
        "my-azure-ai-vision-embeddings": {
            "type": "embedding",
            "endpoint": "http://{adapter-openai-hostname}/openai/deployments/my-azure-ai-vision-embeddings/embeddings",
            "pricing": {
                "unit": "token",
                "prompt": "0.0009"
            },
            "upstreams": [
                {
                    "endpoint": "https://{azure-ai-vision-service-name}.cognitiveservices.azure.com"
                }
            ]
        }
    }
}
```

### OpenAI adapter configuration

```
AZURE_AI_VISION_DEPLOYMENTS=my-azure-ai-vision-embeddings
```


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-162.nightly-test.deltixhub.io
>E2E tests status: failed

adubovik:
/deploy-review
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/531664)

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-162.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/embeddings/azure_ai_vision.py
```diff
@@ -0,0 +1,153 @@
...
+    async def _get_embedding(input: str | Resource) -> VectorizeResponse:
+        if isinstance(input, str):
+            return await _get_text_embedding(creds, endpoint, input)
+        elif isinstance(input, Resource):
+            return await _get_image_embedding(creds, endpoint, input)
+        else:
+            assert_never(input)
+
+    tasks: List[asyncio.Task[VectorizeResponse]] = [
+        asyncio.create_task(_get_embedding(input)) for input in inputs
```
roman-romanov-o: Minor:
`input` is reserved python keyword, let's avoid using such names in any variables

aidial_adapter_openai/embeddings/azure_ai_vision.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+
+    resp.raise_for_status()
+    return VectorizeResponse.parse_obj(resp.json())
+
+
+async def _get_text_embedding(
+    creds: OpenAICreds, endpoint: str, text: str
+) -> VectorizeResponse:
+    resp = await get_http_client().post(
```
roman-romanov-o: Currently we have both `aiohttp` (for `dalle3` and `gpt 4 multi modal`) and `httpx` as http clients 

Also, look like `httpx` still have performance issue owing to `anyio.Lock` :(
https://github.com/encode/httpx/issues/3215

I think we can use aiohttp session here for that reason

aidial_adapter_openai/embeddings/azure_ai_vision.py
```diff
@@ -0,0 +1,153 @@
...
+    usage = Usage(prompt_tokens=n, total_tokens=n)
+
+    return EmbeddingResponse(model=deployment, data=vectors, usage=usage)
+
+
+async def _get_image_embedding(
+    creds: OpenAICreds, endpoint: str, resource: Resource
+) -> VectorizeResponse:
+    resp = await get_http_client().post(
+        url=endpoint.rstrip("/") + "/computervision/retrieval:vectorizeImage",
```
roman-romanov-o, line 140: Since that's upstream endpoint, maybe let's pass the whole endpoint into upstream config, with `/computervision/...`?

aidial_adapter_openai/embeddings/azure_ai_vision.py
```diff
@@ -0,0 +1,153 @@
...
+        content=resource.data,
+        headers={
+            **_get_auth_headers(creds),
+            "content-type": resource.type,
+        },
+        params=_VERSION_PARAMS,
```
roman-romanov-o, line 141: User definitely can pass `api_version` by himself, if some new fields will be introduced into api response
Maybe let's add this `_VERSION_PARAMS` as defaults, if nothing passed?

And if something is passed, we validate that `api-version > 2024-02-01` and `model-version > 2022-04-11`

pyproject.toml
```diff
@@ -24,6 +24,8 @@ tiktoken = "0.7.0"
 uvicorn = "0.23"
 wrapt = "^1.15.0"
 pydantic = "^1.10.12"
+httpx = "^0.27.0"
```
roman-romanov-o, line 27: Why do we need to add this?
`httpx` is included inside of `fastapi`, do we want this specific version?

adubovik: Because it's an explicit dependency of this project. 
It's literally used somewhere as `import httpx`.

aidial_adapter_openai/embeddings/azure_ai_vision.py
adubovik: api-version of the `/embeddings` endpoint doesn't have anything to do with api-version of the Azure AI Vision service. It's apples and oranges.

It's a bad user experience if one needs to go to Azure docs and find out what are the correct API versions for this particular service.

What we could do is to configure the default api version in the env var: `AZURE_AI_VISION_API_VERSION`, as it's been done for `DALLE3_AZURE_API_VERSION`. But I doesn't see any sense to introduce it while we still have a **single API version**.

aidial_adapter_openai/embeddings/azure_ai_vision.py
adubovik: We need to combine `:vectorizeImage` and `:vectorizeText` endpoints into a single endpoint.

The valid options are one of the common prefixes of these two endpoints:

```
SERVICE_HOSTNAME/computervision/retrieval:vectorize
SERVICE_HOSTNAME/computervision/retrieval
SERVICE_HOSTNAME/computervision/
SERVICE_HOSTNAME
```

Since we could not use an endpoint which will actually be called, it's strange for me to ask the user to provide any half-baked URL, which is not documented in any way in the Azure docs.

The only reasonable and simplest option - is to provide the URL of the service itself as it's referenced under the name VISION_ENDPOINT in all of the docs about Image Analysis e.g. this [one](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/quickstarts-sdk/image-analysis-client-library-40?tabs=visual-studio%2Clinux&pivots=programming-language-python).

aidial_adapter_openai/embeddings/azure_ai_vision.py
adubovik: It's not a reserved keyword. It's a [built-in function](https://docs.python.org/3/library/functions.html#built-in-functions).

Not sure what's wrong with using one of those names.
I see it all the time: id, hash, list, dir are very typical variable names.

roman-romanov-o: Wrong terminology, yes

And I also see that sometimes people use `id`/`list` as variable names, but I don't think it's good practice

It's

1. Confuses your IDE (minor stuff)
2. Can lead to a potential bugs, if you'll end up with something like this

```python
list = []
...
# (lots of code, and you forgot that you overshadowed something)

return list(some_generator)
```

I don't understand, why flake actually allows this, and do not throw something like this `F811: redefinition of unused name`

adubovik: 1. It doesn't confuse IDE. IDE resolves symbols by going through scopes - from local to global the same way Python interpreter does it.
2. The example program won't be type-checked. This is perfectly fine Python code. I've renamed the variable to `input_`. I created the issue for the linter: https://github.com/epam/ai-dial-adapter-openai/issues/164

aidial_adapter_openai/embeddings/azure_ai_vision.py
adubovik: Indeed, migrated to `aiohttpx`

aidial_adapter_openai/embeddings/azure_ai_vision.py
roman-romanov-o: 1. I didn't mean code analysis. For example, IDEs are often configured to use different colour pallets for `user-defined functions / variables / built-in functions`. And those colour pallets are not so intelligent to distinguish cases, when you redefined built-ins.
2. 👍

aidial_adapter_openai/embeddings/azure_ai_vision.py
roman-romanov-o: Yes, you're right

roman-romanov-o (approved):

PR #170, opened 2024-11-08
fix: supported approximate tokenization of tools and functions

### LINKED ISSUE #29: Correct token calculation for streaming in case of using functions
Currently we don't calculate tokens for functions and tools

adubovik:
Resolves #29


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-170.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/utils/tokenizer.py
```diff
@@ -30,11 +38,39 @@ def __init__(self, model: str) -> None:
...
             ) from e
 
-    def calculate_text_tokens(self, text: str) -> int:
+    def tokenize_text(self, text: str) -> int:
         return len(self.encoding.encode(text))
 
+    def tokenize_response(self, resp: ChatCompletionResponse) -> int:
+        return sum(map(self._tokenize_response_message, resp.messages))
+
+    def _tokenize_object(self, obj: Any) -> int:
```
roman-romanov-o, line 47: Minor:

https://github.com/python/typing/issues/182#issuecomment-1320974824

Looks like we add such typing in our repos:
```python
JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None
```
instead of `Any`

aidial_adapter_openai/utils/chat_completion_response.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any, Iterable, Literal, Self
+
+from aidial_sdk.utils.merge_chunks import merge_chat_completion_chunks
+from pydantic import BaseModel
+
+
+class ChatCompletionResponse(BaseModel):
```
roman-romanov-o, line 7: You can make this class generic, and use types from openai library, so instead of 
`resp: dict` you would get 

```python
class ChatCompletionResponse(BaseModel, Generic[T]):
   resp: T


class ChatComplationBlock(ChatCompletionResponse[ChatCompletionChunk]):
    ...

...
```

aidial_adapter_openai/utils/chat_completion_response.py
```diff
@@ -0,0 +1,51 @@
...
+    @property
+    def usage(self) -> Any | None:
+        return self.resp.get("usage")
+
+    @property
+    def is_empty(self) -> bool:
+        return self.resp == {}
+
+    @property
+    def finish_reasons(self) -> Iterable[Any]:
```
roman-romanov-o, line 20: I think that after using Generic and TypedDict from openai library, it should infer, that 
`finish_reasons -> Iterable[Literal["stop", "length", "tool_calls", "content_filter", "function_call"]]`,
and we can remove this `-> Iterable[Any]:`

aidial_adapter_openai/utils/chat_completion_response.py
```diff
@@ -0,0 +1,51 @@
...
+        for choice in self.resp.get("choices") or []:
+            if (reason := choice.get("finish_reason")) is not None:
+                yield reason
+
+    @property
+    def has_finish_reason(self) -> bool:
+        return len(list(self.finish_reasons)) > 0
+
+    @property
+    def messages(self) -> Iterable[Any]:
```
roman-romanov-o, line 30: Here I think typing won't be so intelligent, so we would need to extend our generic with two types: both for response, and for choice

aidial_adapter_openai/utils/tokenizer.py
```diff
@@ -43,31 +79,33 @@ def tokens_per_message(self) -> int:
...
 
-    def available_message_tokens(self, max_prompt_tokens: int):
-        return max_prompt_tokens - self.TOKENS_PER_REQUEST
+        tokens += sum(map(self.tokenize_request_message, messages))
+
+        return tokens
 
     @abstractmethod
-    def calculate_message_tokens(self, message: MessageType) -> int:
+    def tokenize_request_message(self, message: MessageType) -> int:
```
roman-romanov-o, line 108: I think it would be easy to confuse `tokenize_request` and `tokenize_request_message` methods

Maybe let's name it just `tokenize_messsage`?

aidial_adapter_openai/utils/tokenizer.py
adubovik: I like the idea of JSON type, but it will lead to **a lot** of additional type checks, the code will be become unreadable.

We should rather migrate to the type provided by the `openai` library, so we don't need to deal with the raw JSON in the first place.

aidial_adapter_openai/utils/chat_completion_response.py
adubovik: I could not use `openai` types just like that, I will have to make a lot of changes in the code downstream.
It's a good idea, let's leave it for another PR.

aidial_adapter_openai/utils/chat_completion_response.py
adubovik: Migrating from JSON to openai types is a different task from the one I'm dealing with in this PR.
Let's leave for another PR.

aidial_adapter_openai/utils/tokenizer.py
adubovik: We need to distinguish response message from a request message, that's why I introduced
`_tokenize_response_message` and `tokenize_request_message`.

`tokenize_message` doesn't help to make the distinction.

roman-romanov-o (approved):

PR #178, opened 2024-11-26
feat: Replace usage of env variables with application config in app state

roman-romanov-o:
The current approach of using env variables makes testing much harder, since variables are in global scope.

Replace it with `app_config` object, that's passed during application init, and could be modified for testing purposes


roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-178.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/app.py
```diff
@@ -68,43 +62,88 @@ async def lifespan(app: FastAPI):
...
 
-app = FastAPI(lifespan=lifespan)
+def create_app(
+    app_config: ApplicationConfig | None = None,
+    to_init_telemetry: bool = True,
+    to_configure_loggers: bool = True,
+) -> FastAPI:
+    app = FastAPI(lifespan=lifespan)
+
+    if app_config is None:
```
adubovik: `app_config = app_config or ApplicationConfig.from_env()`

aidial_adapter_openai/app_config.py
```diff
@@ -0,0 +1,39 @@
+from typing import Dict, List
+
+from pydantic import BaseModel, Field
+
+import aidial_adapter_openai.env as env
+
+
+class ApplicationConfig(BaseModel):
+    MODEL_ALIASES: Dict[str, str] = Field(default_factory=dict)
```
adubovik: You don't need to use `Field` with pydantic. Is will work as well:

```python
MODEL_ALIASES: Dict[str, str] = {}
...
DALLE3_AZURE_API_VERSION: str = "2024-02-01"
```

aidial_adapter_openai/app.py
```diff
@@ -68,43 +62,88 @@ async def lifespan(app: FastAPI):
...
+        )
+    return image_tokenizer
+
+
+app = create_app()
+
+
 @app.post("/openai/deployments/{deployment_id:path}/chat/completions")
-async def chat_completion(deployment_id: str, request: Request):
+async def chat_completion(
```
adubovik: Maybe add the routes inside `create_app()`?

Without it `create_app` doesn't ease the testing much - I couldn't do with it anything.

Besides, if you add the routes inside `create_app()` you don't need to do the config injection via `Depends`.

aidial_adapter_openai/utils/image_tokenizer.py
```diff
@@ -58,14 +54,18 @@ def _compute_high_detail_tokens(self, width: int, height: int) -> int:
...
-    (GPT4O_MINI_IMAGE_TOKENIZER, GPT4O_MINI_DEPLOYMENTS),
-    (GPT4_VISION_IMAGE_TOKENIZER, GPT4_VISION_DEPLOYMENTS),
-]
-
 
-def get_image_tokenizer(deployment_id: str) -> ImageTokenizer | None:
+def get_image_tokenizer(
+    deployment_id: str, app_config: ApplicationConfig
+) -> ImageTokenizer | None:
+    _TOKENIZERS: List[Tuple[ImageTokenizer, List[str]]] = [
```
adubovik: Not sure about moving _TOKENIZERS inside the method - more line breaks, heavier code, harder to read.

aidial_adapter_openai/app.py
```diff
@@ -129,56 +168,62 @@ async def call_chat_completion(
...
+    if deployment_id in app_config.DALLE3_DEPLOYMENTS:
         storage = create_file_storage("images", request.headers)
         return await dalle3_chat_completion(
             data,
             upstream_endpoint,
             creds,
             is_stream,
             storage,
-            DALLE3_AZURE_API_VERSION,
+            app_config.DALLE3_AZURE_API_VERSION,
```
adubovik: Try to place `app_config` as a first argument instead of the last.

The order of arguments should reflect their scope _(which conveys certain information to the reader)_, rather than the chronology of their appearance in the code _(which is random and therefore of no interest)_.

aidial_adapter_openai/app.py
```diff
@@ -68,43 +62,88 @@ async def lifespan(app: FastAPI):
...
 
 
+def _get_image_tokenizer(
+    deployment_id: str, app_config: ApplicationConfig
+) -> ImageTokenizer:
+    image_tokenizer = get_image_tokenizer(deployment_id, app_config)
+    if not image_tokenizer:
+        raise RuntimeError(
+            f"No image tokenizer found for deployment {deployment_id}"
+        )
```
adubovik: The exception raising could be moved to the `get_image_tokenizer` itself.
No real need for the helper, or at least not in this module.

aidial_adapter_openai/app.py
```diff
@@ -68,43 +62,88 @@ async def lifespan(app: FastAPI):
...
+    if app_config is None:
+        app_config = ApplicationConfig.from_env()
+
+    app.state.app_config = app_config
+
+    if to_init_telemetry:
+        init_telemetry(app, TelemetryConfig())
 
+    if to_configure_loggers:
+        configure_loggers()
```
adubovik: Why do we want to make these actions conditional? They were unconditional before this fix.

aidial_adapter_openai/app_config.py
```diff
@@ -0,0 +1,39 @@
...
+    COMPLETION_DEPLOYMENTS_PROMPT_TEMPLATES: Dict[str, str] = Field(
+        default_factory=dict
+    )
+    DALLE3_AZURE_API_VERSION: str = Field(default="2024-02-01")
+    NON_STREAMING_DEPLOYMENTS: List[str] = Field(default_factory=list)
+
+    @classmethod
+    def from_env(cls) -> "ApplicationConfig":
+        return cls(
+            MODEL_ALIASES=env.MODEL_ALIASES,
```
adubovik: Why not inline definitions from `env` here? since `env` module isn't used anywhere but by this module anyway.

aidial_adapter_openai/app.py
roman-romanov-o: `app_config.DALLE3_AZURE_API_VERSION` represents only `api_version` here

We have quite common function signature across different models:
```python
async def chat_completion(
    data: Any,
    upstream_endpoint: str,
    ...
```

And not all of these handlers have `api_version` parameter

If I'll introduce `api_version` as first parameter for dalle, we'll have different signatures of `chat_completion`

I'd rather prefer not to do this

aidial_adapter_openai/app.py

aidial_adapter_openai/app_config.py
roman-romanov-o: Fixed

aidial_adapter_openai/app.py
roman-romanov-o: That's a good idea, I actually want to use this opportunity to fix another problem:

our `app.py` file is quite big
And it's an entry point for new user, where you usually start walkthrough over new app

---

So I've:

1. Created new module `routers`
2. Moved `chat_completions`,`embeddings`, `health` handlers there
3. Moved exception handler in distinct file `exception_handlers.py`

After that `app.py` looks much cleaner:

```

def create_app(
    app_config: ApplicationConfig | None = None,
    to_init_telemetry: bool = True,
) -> FastAPI:
    app = FastAPI(lifespan=lifespan)
    set_app_config(app, app_config or ApplicationConfig.from_env())

    if to_init_telemetry:
        init_telemetry(app, TelemetryConfig())

    configure_loggers()

    app.get("/health")(routers.health)
    app.post("/openai/deployments/{deployment_id:path}/embeddings")(
        routers.embedding
    )
    app.post("/openai/deployments/{deployment_id:path}/chat/completions")(
        routers.chat_completion
    )
    app.exception_handler(OpenAIError)(openai_exception_handler)
    app.exception_handler(pydantic.ValidationError)(pydantic_exception_handler)
    app.exception_handler(DialException)(dial_exception_handler)

    return app


app = create_app()
```

roman-romanov-o: But I still need to use dependency injection to get app config from request

I've wrapped it into `utils.request` module, so that we have
`set_app_config`, `get_app_config` and `get_request_app_config` utils to work with

aidial_adapter_openai/utils/image_tokenizer.py
roman-romanov-o: Replaced with just plain `if/else`

aidial_adapter_openai/app.py
roman-romanov-o: FIxed

aidial_adapter_openai/app.py
roman-romanov-o: I want to remove telemetry initialization from test application.

Removed `to_configure_loggers`,but left `to_init_telemetry`

aidial_adapter_openai/app_config.py
roman-romanov-o: Good idea
Also moved `_get_eliminate_empty_choices` to `ApplicationConfig`

Removed `env.py` file

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-178.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/routers/chat_completion.py
```diff
@@ -0,0 +1,159 @@
...
+        api_version,
+        tokenizer,
+        app_config.ELIMINATE_EMPTY_CHOICES,
+    )
+
+
+async def chat_completion(
+    deployment_id: str,
+    request: Request,
+    app_config: Annotated[ApplicationConfig, Depends(get_request_app_config)],
```
adubovik: Why not simply do this?

```python
async def chat_completion(deployment_id: str, request: Request):
    app_config = get_request_app_config(request)
    ...
```

aidial_adapter_openai/routers/embeddings.py
```diff
@@ -0,0 +1,41 @@
...
+)
+from aidial_adapter_openai.utils.auth import get_credentials
+from aidial_adapter_openai.utils.parsers import parse_body
+from aidial_adapter_openai.utils.request import (
+    get_api_version,
+    get_request_app_config,
+)
+
+
+async def embedding(
```
adubovik: Wny not this?

```python
async def embedding(deployment_id: str, request: Request):
    app_config = get_request_app_config(request)
    ...
```

aidial_adapter_openai/routers/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .chat_completion import chat_completion
```
adubovik: This folder should be rather called `routes` or `endpoints` instead of `routers`.
A router is a component which groups similar endpoints _(e.g. the one starting the same prefix)_: https://fastapi.tiangolo.com/tutorial/bigger-applications/#an-example-file-structure

Here you are defining handlers for individual endpoints.

tests/conftest.py
```diff
@@ -1,25 +1,29 @@
...
+from aidial_adapter_openai.utils.request import get_app_config
 
 
-@pytest.fixture
-def eliminate_empty_choices():
-    with patch(
-        "aidial_adapter_openai.utils.streaming.ELIMINATE_EMPTY_CHOICES", True
-    ):
-        yield
+@pytest_asyncio.fixture
```
adubovik: `@pytest.fixture` ?

aidial_adapter_openai/app_config.py
```diff
@@ -0,0 +1,78 @@
...
+    DALLE3_DEPLOYMENTS: List[str] = []
+    GPT4_VISION_DEPLOYMENTS: List[str] = []
+    MISTRAL_DEPLOYMENTS: List[str] = []
+    DATABRICKS_DEPLOYMENTS: List[str] = []
+    GPT4O_DEPLOYMENTS: List[str] = []
+    GPT4O_MINI_DEPLOYMENTS: List[str] = []
+    AZURE_AI_VISION_DEPLOYMENTS: List[str] = []
+    API_VERSIONS_MAPPING: Dict[str, str] = {}
+    COMPLETION_DEPLOYMENTS_PROMPT_TEMPLATES: Dict[str, str] = {}
+    DALLE3_AZURE_API_VERSION: str = "2024-02-01"
```
adubovik, line 24: We now have defaults for two variables (`DALLE3_AZURE_API_VERSION` and `ELIMINATE_EMPTY_CHOICES`) defined in two places.

To make it nice and clean we need something like this:

```python
class ApplicationConfig(BaseModel):
    class Config:
        extra = "forbid"

    MODEL_ALIASES: Dict[str, str] = {}
    DALLE3_DEPLOYMENTS: List[str] = []
    GPT4_VISION_DEPLOYMENTS: List[str] = []
    MISTRAL_DEPLOYMENTS: List[str] = []
    DATABRICKS_DEPLOYMENTS: List[str] = []
    GPT4O_DEPLOYMENTS: List[str] = []
    GPT4O_MINI_DEPLOYMENTS: List[str] = []
    AZURE_AI_VISION_DEPLOYMENTS: List[str] = []
    API_VERSIONS_MAPPING: Dict[str, str] = {}
    COMPLETION_DEPLOYMENTS_PROMPT_TEMPLATES: Dict[str, str] = {}
    DALLE3_AZURE_API_VERSION: str = "2024-02-01"
    NON_STREAMING_DEPLOYMENTS: List[str] = []
    ELIMINATE_EMPTY_CHOICES: bool = False

    @classmethod
    def from_env(cls) -> "ApplicationConfig":
        def _list_parser(value: str) -> List[str] | None:
            ...

        def _dict_parser(value: str) -> Dict[str, str] | None:
            ...

        def _str_parser(value: str) -> str | None:
            ...

        def _get_eliminate_empty_choices() -> bool | None:
            ...

        parsers = {
            "MODEL_ALIASES": _list_parser,
            "DALLE3_DEPLOYMENTS": _list_parser,
            "GPT4_VISION_DEPLOYMENTS": _list_parser,
            "MISTRAL_DEPLOYMENTS": _list_parser,
            "DATABRICKS_DEPLOYMENTS": _list_parser,
            "GPT4O_DEPLOYMENTS": _list_parser,
            "GPT4O_MINI_DEPLOYMENTS": _list_parser,
            "AZURE_AI_VISION_DEPLOYMENTS": _list_parser,
            "API_VERSIONS_MAPPING": _dict_parser,
            "COMPLETION_DEPLOYMENTS_PROMPT_TEMPLATES": _dict_parser,
            "DALLE3_AZURE_API_VERSION": _str_parser,
            "NON_STREAMING_DEPLOYMENTS": _list_parser,
        }

        values = {}
        for env_name, parser in parsers.items():
            env_value = os.getenv(env_name)
            if env_value is None or (value := parser(env_value)) is None:
                continue
            values[env_name] = value

        if (eliminate := _get_eliminate_empty_choices()) is not None:
            values["ELIMINATE_EMPTY_CHOICES"] = eliminate

        return cls(**values)
```

aidial_adapter_openai/app.py
adubovik: I misplaced the comment - I meant all the places where a new `app_config` parameter was introduced as the **last** argument (this includes `app_config.ELIMINATE_EMPTY_CHOICES`).

These arguments literally represent the configuration of an **adapter-openai deployment**, they have as broad a scope as possible, so it's a bit strange to place them at the very end of an argument list.

aidial_adapter_openai/app.py
```diff
@@ -68,202 +26,30 @@ async def lifespan(app: FastAPI):
...
-init_telemetry(app, TelemetryConfig())
-configure_loggers()
-
-
-def get_api_version(request: Request):
-    api_version = request.query_params.get("api-version", "")
-    api_version = API_VERSIONS_MAPPING.get(api_version, api_version)
+def create_app(
+    app_config: ApplicationConfig | None = None,
+    to_init_telemetry: bool = True,
```
adubovik: `to_do_something` for a boolean flag is a very unorthodox naming convention.

How about simply `init_telemetry`? You could avoid conflict using qualified imports:

```python
from aidial_sdk.telemetry.init import init_telemetry as sdk_init_telemetry
```

aidial_adapter_openai/utils/image_tokenizer.py
```diff
@@ -58,18 +54,20 @@ def _compute_high_detail_tokens(self, width: int, height: int) -> int:
...
+    if deployment_id in app_config.GPT4O_DEPLOYMENTS:
+        return GPT4O_IMAGE_TOKENIZER
+    elif deployment_id in app_config.GPT4O_MINI_DEPLOYMENTS:
+        return GPT4O_MINI_IMAGE_TOKENIZER
+    elif deployment_id in app_config.GPT4_VISION_DEPLOYMENTS:
+        return GPT4_VISION_IMAGE_TOKENIZER
+    else:
+        raise RuntimeError(
+            f"No image tokenizer found for deployment {deployment_id}"
+        )
```
adubovik: This new `RuntimeError` branch is unreachable, but we don't have a way to prove it statically.

The solution could be an introduction of the deployment classifier which assigns to each deployment the following enum:

```python
def classify_chat_deployment(app_config, deployment_id: str) -> ChatCompletionDeploymentType:
    pass

class ChatCompletionDeploymentType(StrEnum):
    DALLE3 = auto()
    MISTRAL = auto()
    DATABRICKS = auto()
    GPT4_VISION = auto()
    GPT4O = auto()
    GPT4O_MINI = auto()
    GPT_TEXT_ONLY = auto()
```

We could exhaustively match this enum and select an appropriate adapter.

aidial_adapter_openai/routers/chat_completion.py
roman-romanov-o: fixed

aidial_adapter_openai/routers/embeddings.py
roman-romanov-o: fixed

aidial_adapter_openai/routers/__init__.py
roman-romanov-o: fixed, renamed to `endpoints`

tests/conftest.py
roman-romanov-o: fixed

aidial_adapter_openai/app_config.py
roman-romanov-o: Used similar approach with `remove_nones`
```python
   @classmethod
    def from_env(cls) -> "ApplicationConfig":
        def _parse_env_deployments(deployments_key: str) -> List[str] | None:
            return parse_deployment_list(os.getenv(deployments_key)) or None

        def _parse_env_dict(key: str) -> Dict[str, str] | None:
            value = os.getenv(key)
            return json.loads(value) if value else None

        def _parse_eliminate_empty_choices() -> bool | None:
            old_name = "FIX_STREAMING_ISSUES_IN_NEW_API_VERSIONS"
            new_name = "ELIMINATE_EMPTY_CHOICES"

            if old_name in os.environ:
                logger.warning(
                    f"{old_name} environment variable is deprecated. Use {new_name} instead."
                )
                return get_env_bool(old_name)
            elif new_name in os.environ:
                return get_env_bool(new_name)

            return None

        deployment_fields = {
            deployment_key: _parse_env_deployments(deployment_key)
            for deployment_key in (
                "DALLE3_DEPLOYMENTS",
                "GPT4_VISION_DEPLOYMENTS",
                "MISTRAL_DEPLOYMENTS",
                "DATABRICKS_DEPLOYMENTS",
                "GPT4O_DEPLOYMENTS",
                "GPT4O_MINI_DEPLOYMENTS",
                "AZURE_AI_VISION_DEPLOYMENTS",
                "NON_STREAMING_DEPLOYMENTS",
            )
        }
        dict_fields = {
            key: _parse_env_dict(key)
            for key in (
                "MODEL_ALIASES",
                "API_VERSIONS_MAPPING",
                "COMPLETION_DEPLOYMENTS_PROMPT_TEMPLATES",
            )
        }

        return cls(
            **remove_nones(
                {
                    **deployment_fields,
                    **dict_fields,
                    "DALLE3_AZURE_API_VERSION": os.getenv(
                        "DALLE3_AZURE_API_VERSION"
                    ),
                    "ELIMINATE_EMPTY_CHOICES": _parse_eliminate_empty_choices(),
                }
            ),
        )
```

aidial_adapter_openai/app.py

aidial_adapter_openai/utils/image_tokenizer.py
roman-romanov-o: Done, refactored `call_chat_completion` with `match/case`

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-178.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/app_config.py
```diff
@@ -0,0 +1,103 @@
...
+            return ChatCompletionDeploymentType.GPT4O
+        elif deployment_id in self.GPT4O_MINI_DEPLOYMENTS:
+            return ChatCompletionDeploymentType.GPT4O_MINI
+        else:
+            return ChatCompletionDeploymentType.GPT_TEXT_ONLY
+
+    @classmethod
+    def from_env(cls) -> "ApplicationConfig":
+        def _parse_env_deployments(deployments_key: str) -> List[str] | None:
+            return parse_deployment_list(os.getenv(deployments_key)) or None
```
adubovik: `parse_deployment_list` never returns None, but it should when the env var is unset.

roman-romanov-o: Fixed, removed parse_deployment_list from utils, moved all logic here

adubovik (approved):

roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-178.nightly-test.deltixhub.io
>E2E tests status: success

PR #182, opened 2024-12-05
fix: improved error handling

adubovik:
Fixes the following issue: certain exceptions raised during the generation of the chunk stream weren't reported via `error` chunk, which led to the stream being interrupted. These exceptions include: 
1. the internal adapter exceptions (e.g. raised inside the `generate_stream` function),
2. non-standard OpenAI exceptions returned by the upstream

### DIFF
--- a/aidial_adapter_openai/app.py
+++ b/aidial_adapter_openai/app.py
@@ -1,7 +1,5 @@
 from contextlib import asynccontextmanager
 
-import aidial_sdk._errors as sdk_error_handlers
-import pydantic
 from aidial_sdk.exceptions import HTTPException as DialException
 from aidial_sdk.telemetry.init import init_telemetry as sdk_init_telemetry
 from aidial_sdk.telemetry.types import TelemetryConfig
@@ -10,7 +8,7 @@ from openai import OpenAIError
 
 import aidial_adapter_openai.endpoints as endpoints
 from aidial_adapter_openai.app_config import ApplicationConfig
-from aidial_adapter_openai.exception_handlers import openai_exception_handler
+from aidial_adapter_openai.exception_handlers import adapter_exception_handler
 from aidial_adapter_openai.utils.http_client import get_http_client
 from aidial_adapter_openai.utils.log_config import configure_loggers, logger
 from aidial_adapter_openai.utils.request import set_app_config
@@ -42,14 +40,9 @@ def create_app(
     app.post("/openai/deployments/{deployment_id:path}/chat/completions")(
         endpoints.chat_completion
     )
-    app.add_exception_handler(OpenAIError, openai_exception_handler)
-    app.add_exception_handler(
-        pydantic.ValidationError,
-        sdk_error_handlers.pydantic_validation_exception_handler,
-    )
-    app.add_exception_handler(
-        DialException, sdk_error_handlers.dial_exception_handler
-    )
+
+    for exc_class in [OpenAIError, DialException]:
+        app.add_exception_handler(exc_class, adapter_exception_handler)
 
     return app
 
--- a/aidial_adapter_openai/app.py
+++ b/aidial_adapter_openai/app.py
@@ -27,7 +27,7 @@ def create_app(
     app_config: ApplicationConfig | None = None,
     init_telemetry: bool = True,
 ) -> FastAPI:
-    app = FastAPI(lifespan=lifespan, debug=True)
+    app = FastAPI(lifespan=lifespan)
     set_app_config(app, app_config or ApplicationConfig.from_env())
 
     if init_telemetry:
--- a/aidial_adapter_openai/app.py
+++ b/aidial_adapter_openai/app.py
@@ -1,5 +1,6 @@
 from contextlib import asynccontextmanager
 
+import aidial_sdk._errors as sdk_error_handlers
 import pydantic
 from aidial_sdk.exceptions import HTTPException as DialException
 from aidial_sdk.telemetry.init import init_telemetry as sdk_init_telemetry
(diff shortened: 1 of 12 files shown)


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-182.nightly-test.deltixhub.io
>E2E tests status: failed

roman-romanov-o:
/deploy-review
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/546021)

aidial_adapter_openai/app.py
```diff
@@ -1,5 +1,6 @@
 from contextlib import asynccontextmanager
 
+import aidial_sdk._errors as sdk_error_handlers
```
roman-romanov-o: I'd rather not import from protected module, given the fact, that we have ownership over sdk

aidial_adapter_openai/app.py
```diff
@@ -45,9 +42,14 @@ def create_app(
...
-    app.exception_handler(OpenAIError)(openai_exception_handler)
-    app.exception_handler(pydantic.ValidationError)(pydantic_exception_handler)
-    app.exception_handler(DialException)(dial_exception_handler)
+    app.add_exception_handler(OpenAIError, openai_exception_handler)
+    app.add_exception_handler(
+        pydantic.ValidationError,
+        sdk_error_handlers.pydantic_validation_exception_handler,
+    )
+    app.add_exception_handler(
+        DialException, sdk_error_handlers.dial_exception_handler
```
roman-romanov-o: All, that `dial_exception_handler` does, is:
```python
def dial_exception_handler(request: Request, exc: Exception) -> JSONResponse:
    assert isinstance(exc, DIALException)
    return exc.to_fastapi_response()
```

We don't need this assert, FastAPI will do this work for us

So maybe let's just substitute this for lambda?
```python
app.add_exception_handler(
    DialException, lambda _, exception: exception.to_fastapi_response()
)
```

aidial_adapter_openai/exception_handlers.py
```diff
@@ -1,56 +1,81 @@
...
+        return exc
 
-def pydantic_exception_handler(request: Request, exc: pydantic.ValidationError):
-    return pydantic_validation_exception_handler(request, exc)
+    return InternalServerError(str(exc))
 
 
-def dial_exception_handler(request: Request, exc: DialException):
-    return exc.to_fastapi_response()
+def openai_exception_handler(request: Request, exc: Exception) -> Response:
```
roman-romanov-o: Maybe the same approach, as I proposed for DialException:

Replace this function with `assert` inside with just `lambda` right in exception handler:
```python
app.add_exception_handler(
   OpenAIError,
   lambda _, err: to_adapter_exception(err).to_fastapi_response()
)
```
What do you think?

aidial_adapter_openai/utils/adapter_exception.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+    def to_fastapi_response(self) -> FastAPIResponse:
+        return FastAPIResponse(
+            status_code=self.status_code,
+            content=self.content,
+            headers=self.headers,
+        )
+
+    def to_fastapi_exception(self) -> FastAPIException:
```
roman-romanov-o: This method is unused

aidial_adapter_openai/exception_handlers.py
```diff
@@ -1,56 +1,81 @@
...
+        r = exc.response
+        httpx_headers = r.headers
 
-        return Response(
-            content=r.content,
+        # httpx library (used by openai) automatically sets
+        # "Accept-Encoding:gzip,deflate" header in requests to the upstream.
+        # Therefore, we may receive from the upstream gzip-encoded
+        # response along with "Content-Encoding:gzip" header.
+        # We either need to encode the response, or
```
roman-romanov-o, line 28: Minor: we either need to **decode** the response, or remove "Content-Encoding" header

aidial_adapter_openai/exception_handlers.py
```diff
@@ -1,56 +1,81 @@
...
+        # remove the "Content-Encoding" header.
+        if "Content-Encoding" in httpx_headers:
+            del httpx_headers["Content-Encoding"]
+
+        headers = {k.decode(): v.decode() for k, v in httpx_headers.raw}
```
roman-romanov-o: You can use `httpx_headers.items()` method, that will do decoding under the hood, and also will handle case, when there is duplicate headers in response:

```python
...
    def items(self) -> typing.ItemsView[str, str]:
        """
        Return `(key, value)` items of headers. Concatenate headers
        into a single comma separated value when a key occurs multiple times.
        """
        values_dict: dict[str, str] = {}
        for _, key, value in self._list:
            str_key = key.decode(self.encoding)
            str_value = value.decode(self.encoding)
            if str_key in values_dict:
                values_dict[str_key] += f", {str_value}"
            else:
                values_dict[str_key] = str_value
        return values_dict.items()
```

aidial_adapter_openai/app.py
adubovik: fixed, removed the import

aidial_adapter_openai/app.py
adubovik: fixed - only handle openai and dial exceptions using the same exception handler. No need for the pydantic error handler since we don't validate the request using pydantic anyway.

aidial_adapter_openai/exception_handlers.py
adubovik: fixed

aidial_adapter_openai/utils/adapter_exception.py
adubovik: fixed, removed

aidial_adapter_openai/exception_handlers.py
adubovik: No, it's correct. 

We are sending to the client the response from the upstream that was  **automatically** decoded behind the scenes by `httpx`, and we also send the `Content-Encoding:gzip` header which was received from the upstream along with the original response.

So we end up with the response content and the header that contradict each other. 

To make them consistent, we need to _EITHER_ **encode** the response _(cancelling out the decoding which httpx did for us)_ using gzip _OR_ remove the "Content-Encoding" header.

aidial_adapter_openai/exception_handlers.py
adubovik: nice, fixed

adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-182.nightly-test.deltixhub.io
>E2E tests status: success

roman-romanov-o (approved):

Issue #179, opened 2024-11-27
Validation for `custom_content` is missing from gpt4 handler


roman-romanov-o:
If you pass `custom_content` (image attachments) and image content parts into `gpt-4` handler, no error will be raised, image will be just ignored.
That's bad UX.

adubovik:
Be careful to check only the DIAL extensions to the OpenAI API - e.g. `custom_content` and image content parts with **DIAL URLs**.

The image content parts with regular URLs may be just fine, since we really don't know the capabilities of the model, which fall into the **catch-all-gpt4** category. It may be GPT-4.5 that supports all kinds of yet unknown modalities, and we should try to proxy the upstream as faithfully as possible. So that even older versions of `adapter-openai` could be used as proxies for the bleeding edge OpenAI models...

PR #183, opened 2024-12-05
fix: tolerate tokenizer failures

adubovik:
Also, improved error message for an unexpected non-textual content part.

## Motivation

The computation of the ` usage` field during streaming is the feature of the Adapter itself.

It's frustrating for an end user to see the Adapter failing on a **successful** response from the upstream only because the Adapter hasn't managed to tokenize the request/response for whatever reason _(the adapter may have a bug in the tokenization logic or the deployment was misconfigured)_.

A safer option is to allow the response to succeed without any `usage` reported, and log the error message in the Adapter.

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -506,6 +506,52 @@ async def test_connection_error_from_upstream(test_app: httpx.AsyncClient):
 
 @respx.mock
 @pytest.mark.asyncio
+async def test_unexpected_multi_modal_input_streaming(
+    test_app: httpx.AsyncClient,
+):
+    mock_stream = OpenAIStream(
+        single_choice_chunk(delta={"role": "assistant"}),
+        single_choice_chunk(delta={"content": "Test response"}),
+        single_choice_chunk(delta={}, finish_reason="stop"),
+    )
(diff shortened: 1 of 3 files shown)


tests/test_errors.py
```diff
@@ -505,6 +505,52 @@ async def test_connection_error_from_upstream(test_app: httpx.AsyncClient):
...
+                }
+            ],
+        },
+        headers={
+            "X-UPSTREAM-KEY": "TEST_API_KEY",
+            "X-UPSTREAM-ENDPOINT": "http://localhost:5001/openai/deployments/gpt-4/chat/completions",
+        },
+    )
+
+    assert response.status_code == 200
```
roman-romanov-o, line 550: Can you pls add todo/issue, that we should return 400 for such requests?

roman-romanov-o (approved):

tests/test_errors.py
roman-romanov-o: Ah, I've forgot, that I created this issue earlier
https://github.com/epam/ai-dial-adapter-openai/issues/179

PR #185, opened 2024-12-09
feat: Add integration tests

roman-romanov-o:
- Added integration test deployments config, to pass right upstreams. Look at `integration_test_config_example.json`
- In addition to `TestCase` dataclass, that was introduced in other integration tests, added `TestSuite` dataclass, that allows to decompose one big test into smaller ones. Look at it in `base.py` file
- Introduced `TestSuiteBuilder` function type, that takes `TestSuite` object, and create new test cases, using `t.test_case(...)` shortcut
- added such builders into `chat_completion_suites` module


roman-romanov-o:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-185.nightly-test.deltixhub.io
>E2E tests status: success

aidial_adapter_openai/app_config.py
```diff
@@ -25,23 +25,31 @@ class ApplicationConfig(BaseModel):
...
-        elif deployment_id in self.DATABRICKS_DEPLOYMENTS:
-            return ChatCompletionDeploymentType.DATABRICKS
-        elif deployment_id in self.GPT4O_DEPLOYMENTS:
-            return ChatCompletionDeploymentType.GPT4O
-        elif deployment_id in self.GPT4O_MINI_DEPLOYMENTS:
-            return ChatCompletionDeploymentType.GPT4O_MINI
-        else:
-            return ChatCompletionDeploymentType.GPT_TEXT_ONLY
+        for deployment_type, attr_name in self.DEPLOYMENT_TYPE_MAP.items():
+            if deployment_id in getattr(self, attr_name):
```
adubovik: Avoid dynamic `getattr` at all costs, it makes the code much more fragile and error-prone.

Now you need to keep in sync the values in `DEPLOYMENT_TYPE_MAP` and the fields in `ApplicationConfig`. The IDE won't help with it, it should be done manually. The type checker won't complain that there is a typo in a value of `DEPLOYMENT_TYPE_MAP` dictionary. It won't complain either if you use a field which has non-list type, e.g. `add_deployment` - it will result in a run-time error.

Such a mistake wasn't possible before due to the type checks.

A more robust type-safe approach would be to replace attributes with the getters:

```python

    DEPLOYMENT_TYPE_MAP: Dict[ChatCompletionDeploymentType, Callable[['ApplicationConfig'], List[str]]] = {
        ChatCompletionDeploymentType.DALLE3: lambda c: c.DALLE3_DEPLOYMENTS,
        ChatCompletionDeploymentType.GPT4_VISION: lambda c: c.GPT4_VISION_DEPLOYMENTS,
    }
```

pyproject.toml
```diff
@@ -37,6 +37,7 @@ aidial-sdk = {version = "^0.16.0", extras = ["telemetry"]}
 pytest = "7.4.0"
 pytest-asyncio = "0.21.1"
 respx = "^0.21.1"
+pytest-xdist = "^3.5.0"
```
adubovik, line 37: let's group `pytest*` packages together

pyproject.toml
```diff
@@ -51,6 +52,7 @@ nox = "^2023.4.22"
 python-dotenv = "^1.0.1"
 
 [tool.pytest.ini_options]
+addopts="-n=auto"
```
adubovik: `addopts = "-n=auto --asyncio-mode=auto"`

You could remove all `@pytest.mark.asyncio` and replace `@pytest_asyncio.fixture` with `@pytest.fixture` when this option is enabled.

tests/integration_tests/base.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def deployment_type(self) -> ChatCompletionDeploymentType:
+        return self.deployment_config.deployment_type
+
+    @classmethod
+    def create(
+        cls,
+        deployment_config: DeploymentConfig,
+        streaming: bool,
+        case_builder: Callable[["TestSuite"], None],
```
adubovik: `: TestSuiteBuilder`

tests/integration_tests/chat_completion_suites/text.py
```diff
@@ -0,0 +1,105 @@
...
+
+    s.test_case(
+        name="pinocchio in one token",
+        max_tokens=1,
+        messages=[user("tell me the full story of Pinocchio")],
+        expected=lambda s: len(s.content.split()) <= 1,
+    )
+
+
+# Mistral do not work properly with stop sequence
```
adubovik: does not

tests/integration_tests/chat_completion_suites/text.py
```diff
@@ -0,0 +1,105 @@
...
+@exclude_deployments(deployment_types=[ChatCompletionDeploymentType.MISTRAL])
+def text_mistral_excluded(s: TestSuite) -> None:
```
adubovik: `test_stop_sequence`

tests/integration_tests/chat_completion_suites/text.py
```diff
@@ -0,0 +1,105 @@
...
+        name="stop sequence",
+        stop=["John", "john"],
+        messages=[user('Reply with "Hello John"')],
+        expected=lambda s: "john" not in s.content.lower(),
+    )
+
+
+# Databricks do not allow consecutive system messages
+@exclude_deployments(deployment_types=[ChatCompletionDeploymentType.DATABRICKS])
+def text_databricks_excluded(s: TestSuite) -> None:
```
adubovik: `test_multi_system_messages`

tests/integration_tests/chat_completion_suites/text.py
```diff
@@ -0,0 +1,105 @@
...
+def text_mistral_excluded(s: TestSuite) -> None:
+    s.test_case(
+        name="stop sequence",
+        stop=["John", "john"],
+        messages=[user('Reply with "Hello John"')],
+        expected=lambda s: "john" not in s.content.lower(),
+    )
+
+
+# Databricks do not allow consecutive system messages
```
adubovik: 1. `does not`
2. does it allow non-consecutive multiple system messages?

tests/integration_tests/chat_completion_suites/text.py
```diff
@@ -0,0 +1,105 @@
...
+    s.test_case(
+        name="pinocchio in one token",
+        max_tokens=1,
+        messages=[user("tell me the full story of Pinocchio")],
+        expected=lambda s: len(s.content.split()) <= 1,
+    )
+
+
+# Mistral do not work properly with stop sequence
+@exclude_deployments(deployment_types=[ChatCompletionDeploymentType.MISTRAL])
```
adubovik: Worth adding this as a valid `ExpectedException` test if it doesn't support this feature, similarly how it's done for various exceptions in `test_common`.

tests/integration_tests/chat_completion_suites/vision.py
```diff
@@ -0,0 +1,30 @@
...
+        ChatCompletionDeploymentType.GPT4O,
+        ChatCompletionDeploymentType.GPT4O_MINI,
+        ChatCompletionDeploymentType.GPT4_VISION,
+    ]
+)
+def vision_common(s: TestSuite) -> None:
+    s.test_case(
+        name="image_in_content_parts",
+        messages=[
+            user_with_image_url("What is this image?", SAMPLE_DOG_RESOURCE),
```
adubovik: Maybe worth querying `Describe the image in one word` to save completion tokens + max_tokens=10 or so.

tests/integration_tests/constants.py
```diff
@@ -0,0 +1,10 @@
+from pathlib import Path
+
+from aidial_adapter_openai.utils.resource import Resource
+
+CURRENT_DIR = Path(__file__).parent
```
adubovik, line 8: Maybe move `TEST_DEPLOYMENTS_CONFIG_PATH` here too?

tests/integration_tests/test_chat_completion.py
```diff
@@ -0,0 +1,90 @@
...
+
+    return [
+        test_case
+        for streaming in (False, True)
+        # for streaming in (True, False)
+        for deployment in TestDeployments.from_config(
+            TEST_DEPLOYMENTS_CONFIG_PATH
+        ).deployments
+        for test_case in TestSuite.create(
+            deployment, streaming, test_case_builder
```
adubovik: This looks overly twisted. Maybe this?

```python
def create_test_cases(builders: List[TestSuiteBuilder]):
    for streaming in (False, True):
        for deployment in TestDeployments.from_config(
            TEST_DEPLOYMENTS_CONFIG_PATH
        ).deployments:
            test_suite = TestSuite(deployment, streaming)
            for builder in builders:
                builder(test_suite)
            yield from test_suite


@pytest.mark.parametrize(
    "test_case",
    create_test_cases(
        [
            text_common,
            text_mistral_excluded,
            text_databricks_excluded,
            tools_common,
            vision_common,
        ]
    ),
    ids=lambda tc: tc.get_id(),
)
    ...
```

tests/integration_tests/test_chat_completion.py
```diff
@@ -0,0 +1,90 @@
...
+
+def create_test_cases(
+    test_case_builder: TestSuiteBuilder,
+) -> List[TestCase]:
+
+    return [
+        test_case
+        for streaming in (False, True)
+        # for streaming in (True, False)
+        for deployment in TestDeployments.from_config(
```
adubovik: `TestDeployments.from_config(TEST_DEPLOYMENTS_CONFIG_PATH)` is repeated twice in the codebase. Maybe it is worth declaring it as a helper?

tests/conftest.py
```diff
@@ -1,22 +1,43 @@
...
 from aidial_adapter_openai.utils.request import get_app_config
+from tests.integration_tests.base import DeploymentConfig, TestDeployments
+
+TEST_DEPLOYMENTS_CONFIG_PATH = os.getenv(
+    "INTEGRATION_TEST_DEPLOYMENTS_CONFIG_PATH",
+    "tests/integration_tests/integration_test_config.json",
+)
 
 
 @pytest.fixture
```
adubovik: `@pytest.fixture(scope="session")` since the instance is kind of a singleton.

tests/utils/openai.py
```diff
@@ -0,0 +1,290 @@
...
+
+def tool_response(id: str, content: str) -> ChatCompletionToolMessageParam:
+    return {"role": "tool", "tool_call_id": id, "content": content}
+
+
+def function_to_tool(function: FunctionDefinition) -> ChatCompletionToolParam:
+    return {"type": "function", "function": function}
+
+
+def sanitize_test_name(name: str) -> str:
```
adubovik: dead code

tests/integration_tests/integration_test_config_example.json
```diff
@@ -0,0 +1,38 @@
+[
```
adubovik: 1/ Could you make it compatible with the DIAL Core config?

As of now, it's very hard for me too configure the config - I need to do it manually for each deployment I want to test.

But if it's compatible with the DIAL Core Config, I could copy-paste it, and that's it!
So the user may configure the config like this:

```json
{
  "models": {
    "gpt-35-turbo-0613": {
      "upstreams": [
        {
          "endpoint": "https://openai-azure-service-name-1.openai.azure.com/openai/deployments/gpt-35-turbo-0613/chat/completions",
          "key": "api-key-1"
        },
        {
          "endpoint": "https://openai-azure-service-name-2.openai.azure.com/openai/deployments/gpt-35-turbo-0613/chat/completions",
          "key": "api-key-2"
        }
      ]
    }
  }
}
```

2/ Then the actual deployments which are going to be tested could be picked up from the environment variables. E.g.

```
GPT4O_DEPLOYMENTS=gpt-4o,gpt-4o-2024-05-13,gpt-4o-2024-08-06
GPT4_VISION_DEPLOYMENTS=gpt-4-vision-preview
DALLE3_DEPLOYMENTS=dall-e-3
```

And we could introduce a special env var for GPT_TEXT_ONLY deployments - `GPT_TEXT_ONLY_DEPLOYMENT`.

We should probably name it `GENERIC_CHAT_COMPLETION_DEPLOYMENTS`...

```
GPT4O_DEPLOYMENTS=gpt-4o,gpt-4o-2024-05-13,gpt-4o-2024-08-06
GPT4_VISION_DEPLOYMENTS=gpt-4-vision-preview
GENERIC_CHAT_COMPLETION_DEPLOYMENTS=gpt-35-turbo-0613
```

The integration tests will be created against these deployments.
And once again - it's very easy to configure - just copy env vars from the environment of the `adapter-openai` deployment.

tests/integration_tests/base.py
```diff
@@ -0,0 +1,194 @@
...
+    n: int | NotGiven
+
+    functions: List[Function] | NotGiven
+    tools: List[ChatCompletionToolParam] | NotGiven
+    temperature: float | NotGiven
+
+    def get_id(self):
+        parts = [
+            sanitize_id_part(self.name),
+            f"{sanitize_id_part(self.deployment_config.deployment_id)}",
```
adubovik: How to distinguish test_case with the same deployment id but **different upstreams**?
Need to print an index of the deployment in the original config to disambiguate the deployments.

aidial_adapter_openai/app_config.py
roman-romanov-o: Good idea, fixed

pyproject.toml
roman-romanov-o: fixed

pyproject.toml
roman-romanov-o: Good idea, fixed

tests/integration_tests/base.py
roman-romanov-o: Fixed

tests/integration_tests/chat_completion_suites/text.py
roman-romanov-o: fixed

tests/integration_tests/chat_completion_suites/text.py
roman-romanov-o: Fixed

tests/integration_tests/chat_completion_suites/text.py
roman-romanov-o: Fixed

tests/integration_tests/chat_completion_suites/text.py
roman-romanov-o: Fixed
No, Databricks does not allow multiple system messages

Changed the comment

tests/integration_tests/chat_completion_suites/text.py
roman-romanov-o: By "does not work properly" I meant that Mistral just ignores this parameter, and returns regular output without stopping

Changed comment to "Mistral just ignores stop sequence"

tests/integration_tests/chat_completion_suites/vision.py
roman-romanov-o: Good idea, done

tests/integration_tests/integration_test_config_example.json
roman-romanov-o: Yeah, nice idea

I've reworked test config in core-config manner

As for `GENERIC_CHAT_COMPLETION_DEPLOYMENTS` - is think that's easier to just stick to logic we're using right now: all unspecified deployments fall back to `GPT_TEXT_ONLY`

But to avoid mistakes, when somebody just forgot to include deployment into needed env variable, I've added information about deployment type to test id

Now it looks like this:

```
test_chat_completion[image_in_custom_content/GPT4O__gpt_4o_2024_05_13/stream:on/tokens:800]
```

tests/integration_tests/base.py
roman-romanov-o: Yeah, I've included upstream index into test name, now it looks like that:
```
test_chat_completion[image_in_custom_content/GPT4_VISION__gpt_4_vision_preview_upstream_1/stream:off/tokens:800]
```

tests/conftest.py
roman-romanov-o: fixed

tests/utils/openai.py
roman-romanov-o: fixed

tests/integration_tests/test_chat_completion.py
roman-romanov-o: I don't see a big need for this, since it's just a one-liner

tests/integration_tests/test_chat_completion.py
roman-romanov-o: Good idea, fixed

tests/integration_tests/constants.py
roman-romanov-o: Done

tests/integration_tests/test_chat_completion.py
```diff
@@ -0,0 +1,102 @@
...
+                    test_case.n,
+                    test_case.functions,
+                    test_case.tools,
+                    test_case.temperature,
+                )
+            except APIError as e:
+                # Somehow, randomly through test, even loop is closing
+                if e.message == "Event loop is closed":
+                    logger.warning("Event loop is closed, retrying...")
+                    continue
```
roman-romanov-o, line 81: That's a bug, that I haven't been able to locate
It's quite random, appears on different tests

So, I decided just to set retries here

aidial_adapter_openai/app_config.py
```diff
@@ -25,23 +25,33 @@ class ApplicationConfig(BaseModel):
...
+                return deployment_type
+        return ChatCompletionDeploymentType.GPT_TEXT_ONLY
+
+    def add_deployment(
+        self, deployment_id: str, deployment_type: ChatCompletionDeploymentType
+    ):
+        if deployment_type == ChatCompletionDeploymentType.GPT_TEXT_ONLY:
+            return
+        config_getter = self.DEPLOYMENT_TYPE_MAP[deployment_type]
+        assert config_getter is not None
```
adubovik: the assert is redundant

tests/integration_tests/chat_completion_suites/text.py
adubovik: I meant that every exclude-test should be paired with an include-test:

```python
@exclude_deployments(deployment_types=[d1, d2])
def test_something(): ...

@include_deployments(deployment_types=[d1, d2])
def test_excluded_something(): ...
```

So that all deployments are covered by the same set of scenarios. This will lead to a code duplication though, so maybe worth combining everything into a single test?

```python
def test_something(): 
    # arrange, act
    # assert
    if deployment in [d1, d2]: ...
    else: ...
```

tests/integration_tests/chat_completion_suites/vision.py
adubovik: `max_tokens=800` looks a bit too much for one word.

tests/integration_tests/test_chat_completion.py
adubovik: It's not about code repetition, but about computing the same thing only once, because 
`TestDeployments.from_config(TEST_DEPLOYMENTS_CONFIG_PATH)` is a test-wide constant, so the right place for it would be alongside `TEST_DEPLOYMENTS_CONFIG_PATH` in `constants.py`:

```python
TEST_DEPLOYMENTS_CONFIG_PATH = ...
TEST_DEPLOYMENT_CONFIG = TestDeployments.from_config(TEST_DEPLOYMENTS_CONFIG_PATH)
```

tests/integration_tests/chat_completion_suites/vision.py
roman-romanov-o: Yeah, forgot that max_tokens do not include prompt tokens

Reduced to `max_tokens=5`

tests/integration_tests/base.py
```diff
@@ -0,0 +1,238 @@
...
+        ]
+
+        if self.max_tokens:
+            parts.append(f"tokens:{sanitize_id_part(self.max_tokens)}")
+        if self.stop:
+            parts.append(f"stop:{sanitize_id_part(self.stop)}")
+        if self.n and self.n != 1:
+            parts.append(f"n:{sanitize_id_part(self.n)}")
+        if self.temperature and self.temperature != 0.0:
+            parts.append(f"temp::{sanitize_id_part(self.temperature)}")
```
adubovik: A single colon? `f"temp:`

aidial_adapter_openai/app_config.py
roman-romanov-o: Removed

tests/integration_tests/test_chat_completion.py
roman-romanov-o: Ok, added it to constants, though it required to add this part

```python
try:
    TEST_DEPLOYMENT_CONFIG = TestDeployments.from_config(
        TEST_DEPLOYMENTS_CONFIG_PATH
    )
except FileNotFoundError:
    TEST_DEPLOYMENT_CONFIG = TestDeployments(
        deployments=[], app_config=ApplicationConfig()
    )
```
To handle cases in CI, where there is no config file in unit tests

tests/integration_tests/chat_completion_suites/text.py
roman-romanov-o: Ah, ok, refactored it with `exclude/include` approach

I want to try to use this instead of one test with `if/else` to see how it will affect readability of tests

tests/integration_tests/chat_completion_suites/tools.py
```diff
@@ -0,0 +1,163 @@
...
+def supports_functions(deployment_type: ChatCompletionDeploymentType):
+    return deployment_type not in [
+        ChatCompletionDeploymentType.DATABRICKS,
+    ]
+
+
+@exclude_deployments(
+    [
+        ChatCompletionDeploymentType.GPT4_VISION,
+        ChatCompletionDeploymentType.DALLE3,
```
adubovik, line 51: Are you sure DALLE3 supports tools?

adubovik (approved):

PR #202, opened 2025-02-24
feat: add configuration for Dall-E-3

adubovik:


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-adapter-openai-pr-202.nightly-test.deltixhub.io
>E2E tests status: success

tests/utils/json.py
```diff
@@ -1,7 +1,7 @@
 from typing import Any
 
 
-def match_objects(expected: Any, actual: Any) -> None:
+def match_objects(expected: Any, actual: Any) -> bool:
```
vladisavvv, line 4: unusual function design, always returns true and checking its result makes no sense. It seems that either the result of the check should be returned or the check should be performed by the function itself (as it was before)

vladisavvv (approved):

tests/utils/json.py
adubovik: I agree, but it makes DX some much more sensible. I simply assert the results of this function instead of simply calling it. Less mental overhead - predicates like this tend to return booleans.

PR #218, opened 2025-05-12
chore: revert: bumps version of sqlite-libs from 3.48.0-r0 to 3.49.1

adubovik:
`python:3.11-alpine` image [depends on](https://hub.docker.com/layers/library/python/3.11-alpine3.21/images/sha256-81d42a73add8e508771ee8e923f9f8392ec1c3d1e482d7d594891d06e78fb51c) `sqlite-libs==3.48.0-r2` which already contains the required [vulnerability fix](https://gitlab.alpinelinux.org/alpine/aports/-/commit/f582ed9d5f50627745fb2fa93498103c53610638).

Therefore, we don't need to upgrade the package ourselves.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,16 +21,14 @@ RUN poetry install --no-interaction --no-ansi --no-cache --only main
 
 FROM python:3.11-alpine as server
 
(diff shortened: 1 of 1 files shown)


tataranovich (approved):
LGTM

Issue #203, opened 2025-02-25
Migrate from aiohttp to httpx


adubovik:
The adapter currently use both `aiohttpx` and `httpx` (indirectly via `openai` library).

It would be great to use only one http client. Otherwise, code uniformity suffers, especially, one of the tests with disparate mocking via `aioresponses` and `respx`.

------

Alternatively, more radically and even better, migrate from making manual HTTP requests to using `openai` library.

PR #233, opened 2025-06-12
fix: enable proxy configuration via env vars for gpt-4o deployments

adubovik:
1. Deployments declared in `GPT4O_DEPLOYMENTS` are called via `aiohttp` library. But default, it doesn't trust_env: https://docs.aiohttp.org/en/stable/client_reference.html
2. Deployments not declared in `GPT4O_DEPLOYMENTS` are called via `httpx` library _(from `openai` lib)_. By default, it does trust_env: https://www.python-httpx.org/environment_variables/

Trusting env means that the client picks up proxy configuration from the env vars _(e.g. `HTTP_PROXY` env var)._

The given fix syncs this feature between the two clients.

In the long run we need to migrate from `aiohttp` altogether: #203

### DIFF
--- a/aidial_adapter_openai/utils/aiohttp.py
+++ b/aidial_adapter_openai/utils/aiohttp.py
@@ -78,7 +78,9 @@ def _get_tracing_timings(trace_request_ctx: dict) -> str:
 @contextlib.asynccontextmanager
 async def post(url: str, headers: Dict[str, str], request: Any):
     ctx = {}
-    async with aiohttp.ClientSession(trace_configs=[_trace_config]) as session:
+    async with aiohttp.ClientSession(
+        trust_env=True, trace_configs=[_trace_config]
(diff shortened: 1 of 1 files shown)


adubovik:
/deploy-review
>Environment: [review-environment](https://chat-ai-dial-adapter-openai-pr-233.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/625445)
>Overlay E2E tests status: [success](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/jobs/1928937)
>Chat E2E tests status: [success](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/jobs/1928936)

sdryapko (approved):

PR #288, opened 2025-09-09
chore: document DIAL Core configuration for various upstream chat APIs

adubovik:


README.md
```diff
@@ -56,6 +56,417 @@ winget install GnuWin32.Make
...
+    }
+  }
+}
+```
+
+</details>
+
+Because the deployment ID is not included in the upstream URL, specify it in the `overrideName` field. If this field is missing, the model name takes the value of the `model` field from the original chat completion request (if present), otherwise `${ADAPTER_DEPLOYMENT_ID}`.
+
+#### OpenAI Platform [Chat Completions API](https://platform.openai.com/docs/api-reference/chat/create)
```
adubovik, line 167: TODO: add an example of a deployment from AI Foundry, e.g. `gpt-oss-120b`

sdryapko (approved):

PR #276, opened 2025-08-21
feat: support file input

adubovik:
Resolves #275 

+ Removed the `Usage` stage and the validation of the supported attachment types, since it requires the adapter to know too much about the models.
+ Note, that file input is only supported in the Responses API for Azure OpenAI models, even though it's possible to pass file inputs in the Chat Completion API.

PR #366, opened 2026-01-21
fix: fix overestimation of prompt tokens for chat completion requests

adubovik:
The PR fixes the bug introduced in https://github.com/epam/ai-dial-adapter-openai/pull/276 in the [release 0.32.0](https://github.com/epam/ai-dial-adapter-openai/releases/tag/0.32.0).

The bug manifests itself in the overestimation of prompt tokens computed by the adapter for chat completion responses.

The overestimation occurs under the following conditions:
1. The requested deployment is registered in either `GPT4O_DEPLOYMENTS` or `GPT4O_MINI_DEPLOYMENTS` variables.
2. The upstream didn't return prompt tokens _(it typically happens when `stream=True` and `include_usage` option isn't provided)_.
3. The request contains an image in a message other than the last one.

When these conditions are met, the adapter incorrectly computes the sum of image tokens over request messages.
So instead of the correct sum:

$$
\sum_{i=0}^{n-1} \text{imageTokens}(\text{message}[i]),
$$

the adapter computes the cumulative sums: 

$$
\sum_{i=0}^{n-1} (n - i) * \text{imageTokens}(\text{message}[i])
$$

This leads to token overestimation that, in the worst-case scenario, is proportional to the number of request messages $n$.

### DIFF
--- a/aidial_adapter_openai/chat_completions/transformation.py
+++ b/aidial_adapter_openai/chat_completions/transformation.py
@@ -10,7 +10,7 @@ from openai.types.chat.chat_completion_assistant_message_param import (
     ContentArrayOfContentPart,
 )
 from openai.types.chat.chat_completion_content_part_param import File
-from pydantic import BaseModel, Field
+from pydantic import BaseModel
 
 from aidial_adapter_openai.dial_api.resource import (
     AttachmentResource,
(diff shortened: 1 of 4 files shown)


adubovik:
/deploy-review

>GitHub actions run: [21218296420](https://github.com/epam/ai-dial-ci/actions/runs/21218296420)

| [deploy-review](https://chat-ai-dial-adapter-openai-pr-366.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/21218296420/job/61045696276) |

aidial_adapter_openai/chat_completions/transformation.py
```diff
@@ -44,15 +44,22 @@ class Error:
...
 
-class ResourceProcessor(BaseModel):
-    class Config:
-        arbitrary_types_allowed = True  # for errors
-
+class MessageTransformer:
     file_storage: FileStorage | None
+    errors: Set[Error]
+    images: List[ImageResource]
+    files: List[FileResource]
```
tg-epam, line 51: looks like simple `list` can be used there

adubovik: Let's modernise the code in a dedicated PR. For now, I would prefer having code with consistent usage of `List`.

tg-epam (approved):

PR #396, opened 2026-02-23
feat: support HEADERS_TO_PROXY env var

adubovik:
Similar to the env var of the same name in the [DIAL adapter](https://github.com/epam/ai-dial-adapter-dial/?tab=readme-ov-file#environment-variables).

Integration with vLLM requires proxying of `x-user-id` header from the client to the vLLM router.
This is a simple fix to achieve this functionality.
However, it seems like we also need to support tokenisation (`max_prompt_tokens` field), which is required for DIAL RAG.
So we should probably create a native integration with vLLM that provides [Tokenizer API](https://docs.vllm.ai/en/stable/serving/openai_compatible_server/#tokenizer-api).

### DIFF
--- a/aidial_adapter_openai/configuration/app_config.py
+++ b/aidial_adapter_openai/configuration/app_config.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import os
-from typing import Dict, List, assert_never
+from typing import Dict, List, Mapping, assert_never
 
 from aidial_sdk.exceptions import InternalServerError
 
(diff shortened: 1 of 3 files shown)


adubovik:
/deploy-review

>GitHub actions run: [22306028235](https://github.com/epam/ai-dial-ci/actions/runs/22306028235)

| [deploy-review](https://chat-ai-dial-adapter-openai-pr-396.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/22306028235/job/64525391441) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/22306028235/job/64526047540) |

adubovik:
The same functionality was implemented in https://github.com/epam/ai-dial-adapter-openai/pull/410

PR #399, opened 2026-03-02
feat: vllm support

sergey-zinchenko:
### Description of changes

Added VLLM deployment support listed by the `VLLM_DEPLOYMENTS` environment variable. 

Deployments listed in `VLLM_DEPLOYMENTS` use proper VLLM tokenization and can be truncated correctly (at least in some cases). Note: The `vllm-openai` image must include two patches for correct adapter functionality:  
1. [Bugfix] Use dedicated MM processor cache in /tokenize to prevent sender-cache pollution [#38545](https://github.com/vllm-project/vllm/pull/38545)  
2. [Bug] Fix Failure in /v1/chat/completions/render for Multimodal Requests [#35684](https://github.com/vllm-project/vllm/pull/35684)  

Ensure the image used includes both fixes to avoid incorrect request routing and tokenization issues.


aidial_adapter_openai/chat_completions/vllm/tokenizer.py
```diff
@@ -0,0 +1,292 @@
+"""
```
adubovik, line 1: It's not general-purpose utils strictly speaking.
Let's move it to `aidial_adapter_openai.chat_completions.vllm.tokenizer` module

aidial_adapter_openai/chat_completions/gpt.py
```diff
@@ -166,3 +193,100 @@ async def get_request_tokens():
...
+    """Pass through streaming chunks from vLLM, injecting
+    ``discarded_messages`` statistics into the last chunk if needed.
+
+    If `eliminate_empty_choices` is True, empty chunk choices are removed
+    (same intent as the standard GPT stream generator).
+    """
+    last_chunk = None
+
+    async for chunk in stream:
+        if eliminate_empty_choices and isinstance(chunk.get("choices"), list):
```
adubovik: This doesn't actually eliminate chunks with an empty list of choices...

I suggest ignoring this flag altogether and simplifying the logic. 
This flag is a historical artefact; it doesn't seem like anyone depends on it.

aidial_adapter_openai/utils/vllm_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+            system_tokens = await self.tokenize_request(
+                original_request, _collect(system_set)
+            )
+            raise TruncatePromptSystemError(max_prompt_tokens, system_tokens)
+
+        kept: Set[int] = set(all_indices)
+
+        def _cascade_remove_tool_replies(start_idx: int) -> None:
+            """Remove consecutive tool replies following *start_idx* and the next assistant."""
+            j = start_idx + 1
```
adubovik: `i` is usually used as a name of an outer loop index.

tests/unit_tests/test_vllm_stream_options.py
```diff
@@ -0,0 +1,168 @@
+from unittest.mock import AsyncMock
+
+import pytest
+
+import aidial_adapter_openai.chat_completions.gpt as gpt_module
+from aidial_adapter_openai.utils.vllm_tokenizer import VllmTokenizer
+
+
+@pytest.mark.asyncio
+async def test_vllm_stream_options_include_usage_injected(monkeypatch):
```
adubovik: Avoid using `monkeypatch` at all costs since it is incredibly fragile.
Please use `respx` to mock responses from tokenize and chat completions endpoints.
See for example `test_stream_options` test.

The same goes for all the tests in the module.

tests/unit_tests/test_vllm_tokenizer.py
```diff
@@ -0,0 +1,601 @@
...
+        messages = [
+            MultiModalMessage(raw_message={"role": "system", "content": "sys"}),
+            MultiModalMessage(raw_message={"role": "user", "content": "huge"}),
+        ]
+
+        with pytest.raises(TruncatePromptSystemAndLastUserError):
+            await tokenizer.truncate_prompt({}, messages, 10)
+
+    @pytest.mark.asyncio
+    async def test_multimodal_messages_sent_as_whole(self):
```
adubovik: The test doesn't test what it claims to test: I can remove `images` and `image_url` from the list of messages, and the test won't fail.

adubovik (changes requested):

README.md
```diff
@@ -633,6 +635,48 @@ The deployment should be added to the environment variable `MISTRAL_DEPLOYMENTS`
...
+```json
+{
+  "models": {
+    "${DIAL_DEPLOYMENT_ID}": {
+      "type": "chat",
+      "overrideName": "${VLLM_MODEL_NAME}",
+      "endpoint": "${ADAPTER_ORIGIN}/openai/deployments/${ADAPTER_DEPLOYMENT_ID}/chat/completions",
+      "upstreams": [
+        {
+          "endpoint": "http://<vllm-host>:<port>/v1/chat/completions"
```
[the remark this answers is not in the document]
sergey-zinchenko: fixed

aidial_adapter_openai/endpoints/chat_completion.py
```diff
@@ -215,6 +230,25 @@ def _get_tokenizer() -> Tokenizer:
 
             return response
 
+        case D.VLLM:
```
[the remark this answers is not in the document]
sergey-zinchenko: done

aidial_adapter_openai/utils/vllm_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+
+class VllmTokenizer:
+    """Tokenizer backed by a remote vLLM ``/tokenize`` endpoint.
+
+    The tokenizer sends the **full** request payload (all messages, tools,
+    etc.) to the vLLM server in a single call and returns the total token
+    count reported by the server.  No per-message, per-modality, or
+    per-attachment token counting is performed on the adapter side.
+
+    This tokenizer **not** performs response token counting.
```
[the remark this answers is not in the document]
sergey-zinchenko: It is about the previous approach: I tried to generate 1 output token to extract usage from the response. It fails because it simply cannot handle a large number of tokens when memory is insufficient, even for a 1-token output. Next, I switched back to that api. Anyway - comment removed.

aidial_adapter_openai/chat_completions/vllm/tokenizer.py
adubovik: `vllm.vllm_tokenizer` is unnecessarily verbose, let keep it simple: `vllm.tokenizer`

aidial_adapter_openai/chat_completions/gpt.py
```diff
@@ -20,20 +20,16 @@
     map_stream,
 )
 from aidial_adapter_openai.utils.tokenizer import Tokenizer
-from aidial_adapter_openai.utils.truncate_prompt import (
-    DiscardedMessages,
```
adubovik, line 24: Let's revert this fix:
1. It has nothing to do with the issue;
2. the type aliases are actually helpful as documentation _(the same way good variable names clarify what the code does)_

aidial_adapter_openai/chat_completions/vllm/tokenizer.py
```diff
@@ -0,0 +1,113 @@
...
+        *,
+        upstream_endpoint: str,
+        extra_headers: dict[str, str] | None = None,
+    ) -> None:
+        self._tokenize_url = derive_tokenize_url(upstream_endpoint)
+        self._extra_headers = extra_headers or {}
+
+        self._http_client = get_http_client()
+
+    async def tokenize(self, request: dict) -> int:
```
adubovik, line 54: Two methods are doing the same thing - `tokenize` and `_call_tokenize`
 Maybe leave just one - `tokenize`?

README.md
```diff
@@ -633,6 +635,47 @@ The deployment should be added to the environment variable `MISTRAL_DEPLOYMENTS`
...
+      ]
+    }
+  }
+}
+```
+
+</details>
+
+Enable the vLLM-specific flow by adding `${ADAPTER_DEPLOYMENT_ID}` to the environment variable `VLLM_CHAT_COMPLETIONS_API_DEPLOYMENTS`.
+##### Request routing / caching headers
```
adubovik: Insert a newline before the header.

aidial_adapter_openai/configuration/app_config.py
```diff
@@ -61,6 +62,8 @@ class ApplicationConfig(ExtraForbidModel):
 
     AUDIO_AZURE_API_VERSION: str = "2025-03-01-preview"
 
+    HEADERS_TO_PROXY: List[str] = []
```
adubovik: I suggest removing this flag until we decide how to solve the session affinity problem in DIAL Core. We can add it later in a separate PR when the design is accepted.

aidial_adapter_openai/chat_completions/vllm/chat_completion.py
```diff
@@ -0,0 +1,148 @@
...
+    *,
+    stream: AsyncIterator[dict],
+    discarded_messages: List[int] | None,
+    eliminate_empty_choices: bool,
+) -> AsyncIterator[dict]:
+    """Pass through vLLM chunks and append discarded message statistics to the last chunk."""
+    last_chunk = None
+
+    async for chunk in stream:
+        if eliminate_empty_choices and isinstance(chunk.get("choices"), list):
```
adubovik: This doesn't actually eliminate chunks with an empty list of choices...

I suggest ignoring this flag altogether and simplifying the logic.
This flag is a historical artefact; it doesn't seem like anyone depends on it.

aidial_adapter_openai/utils/truncate_prompt.py
```diff
@@ -7,52 +25,125 @@
...
+        if raw.get("role") == "assistant" and raw.get("tool_calls"):
+            _cascade_remove_tool_replies(idx)
 
-    discarded_messages = list(set(range(len(messages))) - kept_messages)
+        last_measured_tokens = await tokenizer.tokenize(_build_request(kept))
+        if last_measured_tokens <= max_prompt_tokens:
+            discarded = sorted(all_indices - kept)
+            return _collect(kept), discarded, last_measured_tokens
+
+    if non_system_indices:
```
adubovik: Let's simplify a bit to make it clear which cases lead to `TruncatePromptSystemError` being raised:

```py
    if system_set:
        if kept == system_set:
            system_tokens = last_measured_tokens
        else:
            system_tokens = await tokenizer.tokenize(_build_request(system_set))
        if system_tokens > max_prompt_tokens:
            raise TruncatePromptSystemError(max_prompt_tokens, system_tokens)

    raise TruncatePromptSystemAndLastUserError(
        max_prompt_tokens, last_measured_tokens
    )
```

aidial_adapter_openai/endpoints/chat_completion.py
```diff
@@ -97,8 +106,15 @@ async def call_chat_completion(
     logger.debug(f"deployment api type: {deployment.model_dump_json()}")
     deployment_type, endpoint = deployment.deployment_type, deployment.endpoint
 
-    creds = await get_credentials(request_headers)
-    client = endpoint.get_client({**creds, "api_version": api_version})
+    creds = await get_credentials(
+        request_headers,
+        azure=deployment_type == D.VLLM_CHAT_COMPLETIONS_API,
```
adubovik: `azure=deployment_type != D.VLLM_CHAT_COMPLETIONS_API`

since vLLM doesn't support Azure AD.

aidial_adapter_openai/configuration/app_config.py
```diff
@@ -206,8 +226,10 @@ def from_env(cls) -> ApplicationConfig:
                 "DATABRICKS_DEPLOYMENTS",
                 "GPT4O_DEPLOYMENTS",
                 "GPT4O_MINI_DEPLOYMENTS",
+                "VLLM_CHAT_COMPLETIONS_API_DEPLOYMENTS",
```
adubovik: Let's revert to `VLLM_DEPLOYMENTS`. 
`DeploymentType` doesn't necessarily map one-to-one to deployment env vars.

tests/unit_tests/test_discard_messages.py
```diff
@@ -8,17 +8,13 @@
 )
 from aidial_adapter_openai.utils.multi_modal_message import MultiModalMessage
 from aidial_adapter_openai.utils.tokenizer import Tokenizer
-from aidial_adapter_openai.utils.truncate_prompt import (
```
adubovik, line 11: Revert changes in this file

tests/unit_tests/test_truncate_prompt.py
```diff
@@ -0,0 +1,386 @@
...
+)
+
+from aidial_adapter_openai.utils.multi_modal_message import MultiModalMessage
+from aidial_adapter_openai.utils.truncate_prompt import truncate_prompt
+
+_TEST_IMAGE_URL = "i" * 10
+
+
+def _mm(role: str, content, **extra) -> MultiModalMessage:
+    raw_message = {"role": role, "content": content}
```
adubovik: `raw_message = {"role": role, "content": content, **extra}`

tests/unit_tests/test_truncate_prompt.py
```diff
@@ -0,0 +1,386 @@
...
+
+    tool_call_id = raw.get("tool_call_id")
+    if isinstance(tool_call_id, str):
+        total += len(tool_call_id)
+
+    return total
+
+
+def _messages_char_count(messages: list[MultiModalMessage]) -> int:
+    return sum(_message_len(m) for m in messages)
```
adubovik: If you are testing `truncate_prompt`, which doesn't know anything about multi-modal messages, then there is no sense in all these multi-modal helpers.
The helpers are dealing with the domain which the `truncate_prompt` function isn't aware of.
So you end up testing these test helpers.

```py
def _message(role: str, tokens: int) -> dict:
    return {"raw_message": {"role": role, "tokens": tokens}}

class LengthTokenizerStub:
    def __init__(self, *, call_payloads: list[list[dict]] | None = None):
        self.call_payloads = call_payloads

    async def tokenize(self, request: dict) -> int:
        raw_messages = request.get("messages", [])
        if self.call_payloads is not None:
            # Store a shallow copy to protect against accidental mutation.
            self.call_payloads.append(list(raw_messages))
        # per_request_tokens to represent non-message tokens
        # coming for example from "tools" definitions
        return request["per_request_tokens"] + sum(m["tokens"] for m in raw_messages)


async def _truncate_prompt(per_request_tokens: int, messages: list, max_prompt_tokens: int):
    return await truncate_prompt(
        LengthTokenizerStub(), {"per_request_tokens": 0}, messages, lambda m: m["raw_message"], max_prompt_tokens
    )

...

class TestTruncatePrompt:
    @pytest.mark.asyncio
    async def test_fits_without_truncation(self):
        messages = [_message("system", 5), _message("user", 4)]

        truncated, discarded, used = await _truncate_prompt(0, messages, 9)

        assert discarded == []
        assert used == 9
        assert truncated == messages
```

tests/unit_tests/test_vllm_stream_options.py
```diff
@@ -0,0 +1,196 @@
...
+import respx
+
+from aidial_adapter_openai.utils.request import get_app_config
+from tests.utils.stream import OpenAIStream, single_choice_chunk
+
+_UPSTREAM_ENDPOINT = "http://localhost:5001/v1/chat/completions"
+_API_VERSION = "api-version=2024-12-01-preview"
+
+
+@pytest.fixture
```
adubovik, line 18: Let's simplify:

```py
@pytest.fixture
async def test_app():
    """Configure vllm-test as a vLLM deployment."""
    config = ApplicationConfig().add_deployment(
        "vllm-test", ChatCompletionDeploymentType.VLLM_CHAT_COMPLETIONS_API
    )
    async with create_test_client(config) as client:
        yield client

@respx.mock
@pytest.mark.asyncio
async def test_vllm_stream_options_include_usage_injected(
    test_app: httpx.AsyncClient,
): ...
```

Dockerfile
```diff
@@ -32,8 +32,8 @@ RUN apk upgrade --no-cache libexpat
 RUN pip install setuptools==80.10.2
 # fix CVE-2026-24049
 RUN pip install wheel==0.46.2
-# fix CVE-2025-6965
-RUN apk upgrade --no-cache sqlite-libs
+# fix CVE-2025-6965 and CVE-2026-22184
```
adubovik, line 35: Did Trivy fail without it? In any case, let's do the security fix in a separate PR.

It is much easier to cherry-pick individual commits for a patch release rather than pick individual lines.

adubovik (changes requested):

tests/unit_tests/test_truncate_prompt.py
sergey-zinchenko: It is just a matter of ease of visual understanding of the length (because without extra knowledge, anyone would expect to see a greater token count in messages with images or tool calls, arguments, etc.). Another perspective: we should work with MM messages at the transaction layer, and the truncation layer isn't responsible for how the tokenizer is implemented (i.e., whether it supports or uses the MM part of the messages).

aidial_adapter_openai/endpoints/embeddings.py
```diff
@@ -22,7 +22,11 @@ async def embedding(deployment_id: str, request: Request):
     # See note for /chat/completions endpoint
     request_body["model"] = request_body.get("model") or deployment_id
 
-    creds = await get_credentials(request.headers)
+    creds = await get_credentials(
+        request.headers,
+        azure=deployment_id in app_config.VLLM_DEPLOYMENTS,
```
adubovik: `azure=deployment_id not in app_config.VLLM_DEPLOYMENTS,`

adubovik (changes requested):

adubovik:
/deploy-review

>GitHub actions run: [23552033444](https://github.com/epam/ai-dial-ci/actions/runs/23552033444)

| [deploy-review](https://chat-ai-dial-adapter-openai-pr-399.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/23552033444/job/68568387966) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/23552033444/job/68569575072) |

adubovik (approved):

PR #447, opened 2026-05-06
fix: support chunking_strategy for transcribe model

### LINKED ISSUE #421: Fix issues with the gpt-4o-transcribe-diarize model
[audio_11s.mp3](https://github.com/user-attachments/files/26565895/audio_11s.mp3)
[audio_39s.mp3](https://github.com/user-attachments/files/26565896/audio_39s.mp3)

1. Download the mp3 files and save them to `./attachments` folder
2. Run the adapter via `PORT=5006 make serve`
3. Set env var `SERVICE` pointing to you Azure OpenAI service with `gpt-4o-transcribe-diarize` model
4. Finally, run this script:

```sh
function test_stt() {
  MODEL=${1}
  STREAM=${2}
  FILE=${3}

  URL=http://0.0.0.0:5006/openai/deployments/${MODEL}/chat/completions?api-version=2025-03-01-preview

  data=$(base64 -i ${FILE})

  echo '{"model": "'${MODEL}'", "messages":[{"role":"user", "content": "test", "custom_content": {"attachments": [{"data": "'${data}'", "type": "audio/mpeg"}]}}], "stream": "'${STREAM}'"}' > request.json

  echo "\n===== Testing ${MODEL} with file ${FILE} and stream=${STREAM} =====\n"

  curl -X POST $URL \
    -H "X-UPSTREAM-ENDPOINT:https://${SERVICE}.cognitiveservices.azure.com/openai/deployments/${MODEL}/audio/transcriptions" \
    -H "api-key:dummy-api-key" \
    -d @./request.json
}

test_stt gpt-4o-transcribe-diarize true "./attachments/audio_11s.mp3"
test_stt gpt-4o-transcribe-diarize true "./attachments/audio_39s.mp3"
```

* 11 sec MP3 results an empty output when `stream=True`, and works just fine for `stream=False`
* 39 sec MP3 results in the 400 error `chunking_strategy is required for diarization models` both for `stream=True/False`

AlexFoxalt:
fixes #421


tests/integration_tests/test_audio.py
```diff
@@ -75,6 +86,21 @@ def stt_deployment(request) -> D:
         pytest.skip("No STT deployments were found")
 
 
+if _stt_diarize_deployments:
```
adubovik: Let's use `maybe_parametrized_fixture` decorator.

tests/integration_tests/test_audio.py
```diff
@@ -146,7 +184,60 @@ async def test_speech_to_text(
...
+        create_openai_client(diarize_stt_deployment),
+        stream=stream,
+        deployment_id=diarize_stt_deployment.model_name,
+        messages=[user_with_attachment_data(" ", diarize_audio_resource)],
+    )
+
+    _assert_has_content(response.content)
+
+
+async def test_diarize_default_chunking_auto(
```
adubovik: `test_diarize_not_empty` and `test_diarize_default_chunking_auto` are exactly the same tests

tests/integration_tests/test_audio.py
```diff
@@ -146,7 +184,60 @@ async def test_speech_to_text(
...
         stream=stream,
         deployment_id=stt_deployment.model_name,
-        messages=[message_with_attachment(" ", AUDIO_RESOURCE)],
+        messages=[message_with_attachment(" ", AUDIO_11s_RESOURCE)],
     )
 
     assert is_close_enough(text_query, response.content)
+
+
+async def test_diarize_not_empty(
```
adubovik: `test_diarize_not_empty` tests the same thing as `test_speech_to_text`.
It's better parametrize `test_speech_to_text` with `AUDIO_11s_RESOURCE` and `AUDIO_39s_RESOURCE`.

tests/integration_tests/test_audio.py
```diff
@@ -146,7 +184,60 @@ async def test_speech_to_text(
...
+        create_openai_client(diarize_stt_deployment),
+        stream=stream,
+        deployment_id=diarize_stt_deployment.model_name,
+        messages=[user_with_attachment_data(" ", diarize_audio_resource)],
+    )
+
+    _assert_has_content(response.content)
+
+
+async def test_diarize_long_audio_without_chunking_fails(
```
adubovik, line 191: This is the only test where you need `diarize_stt_deployment`.

All other new tests are redundant, since they repeat what was already in `test_speech_to_text` test. We should rather extend `test_speech_to_text` to accommodate `AUDIO_39s_RESOURCE` then create new tests.

aidial_adapter_openai/audio_api/transcribe/adapter.py
```diff
@@ -128,8 +135,8 @@ async def _handler(request: DIALRequest, response: DIALResponse) -> None:
                         )
 
                     match chunk:
-                        case TranscriptionTextSegmentEvent():
-                            pass
+                        case TranscriptionTextSegmentEvent(text=text):
```
adubovik, line 138: Under which conditions are` TranscriptionTextSegmentEvent` and `TranscriptionTextDeltaEvent` events emitted? What's the difference?
Consult with the API documentation. Could it be that we end up with duplicated text because we channel the content of both events into the same top-level content.

adubovik (changes requested):

tests/integration_tests/test_audio.py
AlexFoxalt: Replaced with `@maybe_parametrized_fixture`

tests/integration_tests/test_audio.py
AlexFoxalt: Added missing `extra_body` field and renamed test case slightly

tests/integration_tests/test_audio.py
AlexFoxalt: Extended `test_speech_to_text` and removed `test_diarize_not_empty` since duplicated

tests/integration_tests/test_audio.py
AlexFoxalt: `test_diarize_long_audio_without_chunking_fails` -> remained
`test_diarize_chunking_auto` -> deleted

aidial_adapter_openai/audio_api/transcribe/adapter.py
AlexFoxalt: According to the [OpenAI docs](https://developers.openai.com/api/reference/resources/audio), the emitted event type depends on the `response_format` value:

* `response_format="diarized_json"` → `TranscriptionTextSegmentEvent`
* `response_format=<other>` → `TranscriptionTextDeltaEvent`

Since these event types are mutually exclusive, we can use a logical OR when handling them. Receiving both types within a single request is not expected.

I also did a quick verification locally:

* `gpt-4o-transcribe` with `response_format="diarized_json"` → not compatible (`400` error)
* `gpt-4o-mini-transcribe` with `response_format="diarized_json"` → not compatible (`400` error)
* `gpt-4o-transcribe-diarize` (default config) → emits only `TranscriptionTextSegmentEvent` + `EventDone`
* `gpt-4o-transcribe` / `gpt-4o-mini-transcribe` with default `response_format="json"` → emit only `TranscriptionTextDeltaEvent` + `EventDone`

adubovik (approved):

PR #492, opened 2026-07-06
fix: preserve static tools results in subsequent calls

### LINKED ISSUE #470: Static tools' results not preserved in subsequent calls
### Name and Version

latest

### What steps will reproduce the bug?

1. Select model with Web Search static tool (e.g. GPT-5.2 with Web Search)
2. Ask `What's the weather in Washington DC on June 10th? Respond with temperature only.`
3. After model responds, ask `Any precipitation?`

### What is the expected behavior?

On second request, model does not call web search, since it was already called on first iteration.

### What do you see instead?

Model calls Web Search two times: on first and second iteration.

### Additional information

Additionally, model chat-completion response contains `responseId`, which looks like a bug:

Image

AlexFoxalt:
fixes #470


aidial_adapter_openai/chat_completions/transformation.py
```diff
@@ -196,19 +197,32 @@ async def download_content(
...
-            "attachments", custom_content.get("attachments") or []
+            "attachments", custom_content.pop("attachments", None) or []
         )
 
+        if "custom_content" in message:
+            if custom_content:
+                message["custom_content"] = custom_content
+            else:
+                message.pop("custom_content")
+
```
adubovik: What's the point of this block? It looks redundant.

aidial_adapter_openai/dial_api/state.py
```diff
@@ -0,0 +1,9 @@
+from openai.types.responses import ResponseFunctionWebSearch
+from pydantic import BaseModel
+
+
+class MessageState(BaseModel):
+    web_search_content: ResponseFunctionWebSearch
```
adubovik: Let's generalize to a list of output items, as we may need to support other output types in the future:

```py
class MessageState(BaseModel):
    responses_output: list[ResponseOutputItem]
```

Note, that we must not blow up on the invalid state! See for a reference

https://github.com/epam/ai-dial-adapter-anthropic/blob/9e2029493742429fa048acaa5e698786e6a28460/aidial_adapter_anthropic/adapter/_claude/state.py#L32-L44

adubovik (changes requested):

aidial_adapter_openai/chat_completions/transformation.py
AlexFoxalt: It removes empty `custom_content` after `attachments` are popped out.
It handles the case where input has `"custom_content": null` (or empty): without this block, that key can remain as `None`/empty in `raw_message`.
It keeps output shape cleaner by omitting `custom_content` when it carries no data.

aidial_adapter_openai/dial_api/state.py
AlexFoxalt: State logic reworked from specific field per object -> field that acts as an objects container

aidial_adapter_openai/chat_completions/transformation.py
```diff
@@ -195,19 +196,32 @@ async def download_content(
     async def transform_message(self, message: dict) -> MultiModalMessage:
         message = ensure_dict("message", message).copy()
 
-        content = ensure_list_or_str("content", message.get("content") or "")
+        raw_content = message.get("content")
```
adubovik: What's the point of introducing `raw_content`? In other words, what was wrong with the one-liner before the fix?

```py
        content = ensure_list_or_str("content", message.get("content") or "")
```

aidial_adapter_openai/chat_completions/transformation.py
adubovik: Why not pop "custom_content" field right away instead of get-then-pop?
That's how it was before the fix.

aidial_adapter_openai/chat_completions/transformation.py
AlexFoxalt: Reworked

aidial_adapter_openai/chat_completions/transformation.py
AlexFoxalt: Reworked

adubovik (approved):

PR #509, opened 2026-07-24
feat: Responses API passthrough: support tokenization endpoint

### LINKED ISSUE #485: Responses API passthrough: support count tokens endpoint
As per [doc](https://developers.openai.com/api/reference/resources/responses/subresources/input_tokens/methods/count)

AlexFoxalt:
fixes #485

### DIFF
--- a/aidial_adapter_openai/endpoints/responses.py
+++ b/aidial_adapter_openai/endpoints/responses.py
@@ -1,4 +1,4 @@
-from collections.abc import AsyncIterator, Mapping
+from collections.abc import AsyncIterator
 from dataclasses import dataclass
 from typing import Self, TypeVar
 
@@ -193,6 +193,16 @@ async def _to_fast_api_response(
 def _to_response_with_headers(
     response: LegacyAPIResponse[_ResponsesPayloadT],
 ) -> ResponseWithHeaders[dict | AsyncIterator[dict]]:
+    response_headers = dict(response.http_response.headers)
+
+    # Reformatting or reading content may invalidate content length.
+    # We don't recompress the response, therefore,
+    # the content encoding may invalidate too.
+    for header in _INVALIDATED_RESPONSE_HEADERS:
+        for key in list(response_headers):
+            if key.lower() == header:
+                del response_headers[key]
+
     parsed_response = response.parse()
 
     if isinstance(parsed_response, AsyncStream):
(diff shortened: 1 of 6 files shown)


aidial_adapter_openai/endpoints/responses.py
```diff
@@ -149,6 +151,29 @@ async def responses_delete(
...
+        await context.client.responses.with_raw_response.input_tokens.count(
+            **request_body
+        )
+    )
+
+    return FastAPIResponse(
+        content=response.http_response.content,
+        headers=_to_safe_response_headers(response.http_response.headers),
+        status_code=response.http_response.status_code,
+    )
```
adubovik: Can we reuse the existing helpers?

```py
    response_with_headers = _to_response_with_headers(response)
    return await _to_fast_api_response(request, response_with_headers)
```

aidial_adapter_openai/endpoints/responses.py
```diff
@@ -164,23 +189,33 @@ async def _to_fast_api_response(
...
         body = _to_dict(parsed_response)
 
-    return ResponseWithHeaders(headers=dict(response_headers), body=body)
+    return ResponseWithHeaders(
+        headers=_to_safe_response_headers(response.http_response.headers),
+        body=body,
+    )
+
+
+def _to_safe_response_headers(
```
adubovik: Could you revert this fix? The idea was that the sanitisation of headers should happen in a single helper, `_to_response_with_headers`, and this helper should be **reused** in all other handlers.

adubovik (changes requested):

aidial_adapter_openai/endpoints/responses.py
AlexFoxalt: Without changes inside these helpers - no
We need to modify their business logic and input signatures since returned response type `InputTokenCountResponse` has different structure then other responses from other endpoints.
I decided to just separate small helper that we really only need.

aidial_adapter_openai/endpoints/responses.py
AlexFoxalt: Depends on [this conversation](https://github.com/epam/ai-dial-adapter-openai/pull/509#discussion_r3644690084)

adubovik (dismissed):

adubovik (approved):

PR #544, opened 2026-09-07
feat: simplify logic behind retrieval of an upstream model name

adubovik:
Unifies how the adapter derives the upstream model name *(i.e. the upstream deployment id)*. Every endpoint used to resolve it on its own, from its own ad-hoc chain of the `X-DIAL-OVERRIDE-NAME` header, the request body `model` field and the `deployment_id` path parameter. Now there is a single rule per endpoint family.

## Derivation of the upstream deployment id

`→` reads as *"otherwise fall back to"*.

|Endpoint|Before|After|
|---|---|---|
|`POST /openai/deployments/{deployment_id}/chat/completions`|body `model` → `deployment_id`|`X-DIAL-OVERRIDE-NAME` → `deployment_id`|
|`POST /openai/deployments/{deployment_id}/embeddings`|body `model` → `deployment_id`|`X-DIAL-OVERRIDE-NAME` → `deployment_id`|
|`POST /openai/deployments/{deployment_id}/tokenize`|`X-DIAL-OVERRIDE-NAME` → body `model` → `deployment_id`|`X-DIAL-OVERRIDE-NAME` → `deployment_id`|
|`POST /openai/deployments/{deployment_id}/truncate_prompt`|`X-DIAL-OVERRIDE-NAME` → `deployment_id`|`X-DIAL-OVERRIDE-NAME` → `deployment_id`|
|`GET /openai/deployments/{deployment_id}/configuration`|`deployment_id`|`X-DIAL-OVERRIDE-NAME` → `deployment_id`|
|`/openai/v1/responses*` *(no `deployment_id` in the URL)*|`X-DIAL-OVERRIDE-NAME` → body `model`|body `model`|
|Anthropic passthrough `/anthropic/*` *(no `deployment_id` in the URL)*|body `model`|body `model`|

## Why each family uses its own source

**Endpoints without a deployment id in the URL: the body `model` field is sufficient, the header adds nothing.** DIAL Core applies the model configuration's `overrideName` to the body `model` field on its way to the adapter, so by the time the request arrives the field already holds the upstream name. Consulting `X-DIAL-OVERRIDE-NAME` on top of it would only re-apply a substitution that has already happened. *(The Anthropic passthrough never consulted the header in the first place — it forwards the request body verbatim.)*

**Endpoints with a deployment id in the URL: the header is required, because the body `model` field cannot be trusted.** Core rewrites that field only when the model configuration declares `overrideName`; without it, `model` is whatever the client put there. Forwarding it upstream — as the adapter used to do — lets a user address an arbitrary deployment of the upstream service through a DIAL model configured for a different one, and have the request dispatched, tokenized and billed as that deployment. So the deployment id is now taken from the adapter's own endpoint URL and the user-supplied `model` is overwritten with it. Since Core leaves the deployment id in the URL untouched, `X-DIAL-OVERRIDE-NAME` — which Core fills in from `overrideName` — is the only channel through which an override can reach these endpoints, hence it takes precedence over the path parameter.

## Behavior changes

- **The request body `model` field is ignored on the Azure-style endpoints.** The deployment id coming from the endpoint always wins, so an arbitrary `model` from a user request can no longer reach the upstream or influence tokenizer and vLLM API type selection.
- **`X-DIAL-OVERRIDE-NAME` now drives deployment dispatch** on every Azure-style endpoint, not just `tokenize` and `truncate_prompt`: the env-var deployment lists (`GPT4O_DEPLOYMENTS`, `DALLE3_DEPLOYMENTS`, `VLLM_DEPLOYMENTS`, …), `TIKTOKEN_MODEL_MAPPING`, `NON_STREAMING_DEPLOYMENTS` and vendor/credential selection are all keyed on the resolved name. **Deployments whose `overrideName` differs from the deployment id in the adapter URL need those env-var lists updated to the `overrideName` value.**
- **`/openai/v1/responses*` no longer reads `X-DIAL-OVERRIDE-NAME`.** It takes the name from the body `model` field, which DIAL Core has already overridden.
- **`GET /openai/deployments/{deployment_id}/configuration` now requires `X-UPSTREAM-ENDPOINT`.** The DIAL Core < 0.41.0 fallback, which guessed the configuration schema from `DALLE3_DEPLOYMENTS`/`GPT_IMAGE_1_DEPLOYMENTS`, is gone and a missing header is an error.


adubovik:
/deploy-review

>GitHub actions run: [34343119210](https://github.com/epam/ai-dial-ci/actions/runs/34343119210)

| [deploy-review](https://chat-ai-dial-adapter-openai-pr-544.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34343119210/job/102438321231) |
| chat | [Failed ❌](https://github.com/epam/ai-dial-ci/actions/runs/34343119210/job/102446201366) |

aidial_adapter_openai/app.py
```diff
@@ -66,6 +66,9 @@ def create_app(
     app.get("/health")(endpoints.health)
 
     app.post("/openai/v1/responses")(endpoints.responses_create)
+    app.post("/openai/v1/responses/input_tokens")(
+        endpoints.responses_input_tokens
+    )
```
AlexFoxalt: It was ok when generation CRUD methods were grouped
Why moved?

alex-chernyshov (changes requested):

aidial_adapter_openai/app.py
adubovik: reverted

adubovik:
/deploy-review

>GitHub actions run: [34349980088](https://github.com/epam/ai-dial-ci/actions/runs/34349980088)

| [deploy-review](https://chat-ai-dial-adapter-openai-pr-544.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34349980088/job/102460577173) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34349980088/job/102465313496) |

alex-chernyshov (approved):

PR #547, opened 2026-09-10
feat!: add smart AWS STS session tags configuration

alex-chernyshov:
## Summary

Generalizes AWS STS session tags ([#539](https://github.com/epam/ai-dial-adapter-openai/pull/539)): the operator picks the tag keys, the requested model joins the DIAL user info as a value source, and the assumed role session is named after the user project.

A single `AWS_SESSION_TAGS` variable declares the tags as a JSON dictionary of `<tag key>: <value source>`:

```ini
AWS_SESSION_TAGS={"application":"Bedrock.modelId","project":"UserInfo.project","employee":"UserInfo.userClaims.email"}
```

The **key** is the name the tag is passed to AWS under — pick whatever your IAM policies expect. The **value source** names where the value comes from:

| Value source | Provides |
|---|---|
| `Bedrock.modelId` | The requested model |
| `UserInfo.<path>` | A field of the DIAL `GET /v1/user/info` response |

Setting the variable enables the feature; unsetting it disables it. Entries whose value source is unknown (`Nope.project`, `Bedrock.region`, an unprefixed `project`) are skipped with a warning.

As before, the tags apply only to the assume-role credentials. The Bedrock bearer token, the static AWS credentials and the AWS credential provider chain are left unchanged.

## Changes

- **`Bedrock.modelId` value source** — the deployment id from the request path with `models[*].overrideName` applied, i.e. the model name actually sent to Bedrock. The tag is skipped with a warning when the request carries no model id, which is the case for the Responses API operations addressed to a stored response rather than to a model *(retrieve, delete, cancel)*.
- **Operator-chosen tag keys** — previously the tag key was the user-info path itself, which forced IAM policies to be written against `roles.0` and `userClaims.email`. The key and the value source are now independent.
- **Independent sources** — a failed `GET /v1/user/info`, a missing `DIAL_URL`, an absent API key, or an unreadable model each cost only their own tags. Failing to retrieve a tag never fails the request: it's logged as a warning and skipped, and the remaining tags are still passed. The user info is fetched only when some tag actually wants it.
- **`RoleSessionName` from the project** — when a tag takes the `UserInfo.project` value source and the user has a project, the session is named `Project_<project>` instead of `BedrockAccessSession`. Independent of the tag key you choose for it; users with no project fall back to the default name.
- **Tag sanitization** — fixes an `AssumeRole` `ValidationError`. Beyond the documented length limits, AWS accepts only letters, digits, spaces and `_ . : / = + - @` in a key or value — a constraint the session-tag docs don't mention, and one that every JSON-serialized value violates through its `,`. Rejected characters are replaced with `_`, one for one so lengths are preserved; keys that collide once truncated or sanitized get a `_1`, `_2`, … postfix rather than being dropped *(previously they were dropped)*. Matching uses Unicode categories, so non-Latin scripts pass through intact.
- **Scope** — the tags reach every Bedrock client the adapter creates. They are never built for the `/anthropic` passthrough, which authenticates to Azure only here.

## ⚠️ Breaking configuration change

`AWS_SESSION_TAGS_FIELDS` is removed and is **silently ignored** — deployments that still set it stop sending tags altogether. The old variable listed `UserInfo` paths, which now go on the right-hand side:

```diff
- AWS_SESSION_TAGS_FIELDS=roles.0,project
+ AWS_SESSION_TAGS={"roles":"UserInfo.roles.0","project":"UserInfo.project"}
```


adubovik (approved):

AlexFoxalt:
/deploy-review

>GitHub actions run: [34471223238](https://github.com/epam/ai-dial-ci/actions/runs/34471223238)

| [deploy-review](https://chat-ai-dial-adapter-openai-pr-547.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34471223238/job/102851320214) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34471223238/job/102853572507) |

